PR #24, opened 2026-04-25
feat(logging): add SecretMaskFilter for masking Authorization headers

### LINKED ISSUE #15: Add reusable SecretMaskFilter for masking Bearer/Token headers in debug logs
## Problem

Every downstream server that talks to an upstream REST API with an API token ends up re-implementing the same log-sanitiser. Currently `fastmcp-pvl-core` exports no logging filter for masking secrets.

## Proposal

Add a `SecretMaskFilter` to `fastmcp_pvl_core` that masks `Authorization: Token …` and `Authorization: Bearer …` values at DEBUG level, so downstream servers can attach it once rather than each reimplementing it.

Suggested API:

```python
from fastmcp_pvl_core import SecretMaskFilter

logger = logging.getLogger(__name__)
logger.addFilter(SecretMaskFilter())
```

The filter should:
- Match `Authorization: Token <value>` and `Authorization: Bearer <value>` (case-insensitive, in both header-style and dict-repr formats)
- Replace the secret with `***`, e.g. `Token ***`
- Pass all other log records through unmodified
- Always return `True` (never suppress records)

## Current workaround

Downstream servers (e.g. `pvliesdonk/paperless-mcp`) are implementing a local `_SecretMaskFilter` in their HTTP client module, pending this upstream addition.

Happy to PR.

pvliesdonk:
Closes #15

## Summary
- Adds `SecretMaskFilter` to `fastmcp_pvl_core`, exported from the top-level package and added to `__all__`.
- Redacts `Authorization: Bearer/Token` values in formatted log messages, so downstream servers can attach it once instead of each reimplementing the same regex (e.g. `paperless-mcp/src/paperless_mcp/client/_http.py`'s local `_SecretMaskFilter`).
- Matches header-style (`Authorization: Bearer xyz`) **and** dict-repr (`'Authorization': 'Token xyz'`) representations, case-insensitive on both the `Authorization` keyword and the `Bearer`/`Token` scheme name. Original scheme casing is preserved in the redacted output.
- Lives in `_logging.py` alongside `configure_logging_from_env`.

## Behavioral contract
- Always returns `True` (redactor, not gatekeeper — never suppresses records).
- For `%`-formatted records, replaces `record.msg` and clears `record.args` so subsequent `getMessage()` calls cannot re-expand the original args and re-introduce the secret.
- Records without a match pass through untouched.
- Catches exceptions from `getMessage()` defensively — a broken format string upstream must not silence the entire log stream.

## Test plan
- [x] 13 new unit tests (`tests/test_logging.py::TestSecretMaskFilter`)
  - Header-style + dict-repr for both `Token` and `Bearer`
  - Case-insensitive matching, scheme casing preserved
  - Multiple occurrences in a single record
  - `%`-format args expansion (with re-call of `getMessage()`)
  - Always-True return for matched / unmatched / broken-format records
  - Negative cases: `X-Api-Key: Token …` is **not** masked, bare "Token holder" is **not** mutated
- [x] Full suite: 209 passed
- [x] `ruff format` / `ruff check` clean
- [x] `mypy --strict` clean

### DIFF
--- a/src/fastmcp_pvl_core/__init__.py
+++ b/src/fastmcp_pvl_core/__init__.py
@@ -23,7 +23,7 @@ from fastmcp_pvl_core._factory import (
     compute_app_domain,
 )
 from fastmcp_pvl_core._icons import IconSpec, make_icon, register_tool_icons
-from fastmcp_pvl_core._logging import configure_logging_from_env
+from fastmcp_pvl_core._logging import SecretMaskFilter, configure_logging_from_env
 from fastmcp_pvl_core._middleware import wire_middleware_stack
 from fastmcp_pvl_core._server_info import (
     UpstreamProvider,
@@ -37,6 +37,7 @@ __all__ = [
     "ArtifactStore",
     "AuthMode",
     "IconSpec",
+    "SecretMaskFilter",
     "ServerConfig",
     "TokenRecord",
     "Transport",
--- a/src/fastmcp_pvl_core/_logging.py
+++ b/src/fastmcp_pvl_core/_logging.py
@@ -53,9 +53,9 @@ def configure_logging_from_env(*, verbose: bool = False) -> None:
 
 
 class SecretMaskFilter(logging.Filter):
-    """Redact ``Authorization: Bearer/Token`` values in log records.
+    """Redact ``Authorization: Bearer/Token/Basic`` values in log records.
 
-    Attach to a logger to mask secret tokens in formatted messages
+    Attach to a logger to mask secret credentials in formatted messages
     before they reach handlers — typically wired up by HTTP-client
     modules that log request/response details at ``DEBUG`` level::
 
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Pushed `ffc0e00` addressing both bot reviews:

**Applied (gemini, security-medium)**
- Added `Basic` to the scheme alternation. Basic auth carries base64-encoded credentials in the same `Authorization` header — same exposure risk, same redaction. New `test_masks_basic_header` covers it; docstring updated.

**Applied (claude-review)**
- Fixed `test_handles_broken_format_string`. The reviewer was right: `args=()` is falsy, so `getMessage()` short-circuits `%`-formatting and never raises — the `except` branch was genuinely uncovered. Switched to `args=("a", "b", "c")` against a single-placeholder format string, which raises `TypeError: not all arguments converted during string formatting`. Verified manually that `getMessage()` raises and `filter()` still returns `True`. Removed `# pragma: no cover` from the `except` block.
- Added `test_masks_equals_separator` to exercise the `=` path the reviewer flagged as a "minor gap" — keeps the regex alternation honest.

211 tests pass; ruff format/check + mypy --strict clean. Resolving the gemini thread.

Issue #21, opened 2026-04-25
MCP File Exchange v0.3: implement FileExchange runtime (env detect, exclusive-create init, atomic IO, lifecycle, consumer resolution)


pvliesdonk:
## Summary

Companion to [#20](https://github.com/pvliesdonk/fastmcp-pvl-core/issues/20). Where #20 ships the protocol surface (typed `FileRef`, `ExchangeURI` parser, capability-declaration helper), this issue ships the **runtime** that lets a server actually participate in the `exchange` transfer method: env-driven group membership, race-safe `.exchange-id` initialisation, atomic write/read of namespaced files, and a consumer-side URI resolver.

A server can adopt v0.3 with `http`-only transport using #20 alone — this issue is what makes the zero-cost shared-volume path work.

**Blocked by:** [pvliesdonk/fastmcp-pvl-core#20](https://github.com/pvliesdonk/fastmcp-pvl-core/issues/20)

### `FileExchange` class

The single entry point. One instance per server. Constructed eagerly from env vars; falls back to a sentinel \"not configured\" state if env is missing so callers don't need to special-case.

```python
from fastmcp_pvl_core import FileExchange

fx = FileExchange.from_env(default_namespace=\"image-mcp\")

if fx.is_configured:
    uri = fx.write_atomic(
        origin_id=\"a1b2c3\",       # MUST be validated as raw JSON, not URI-decoded
        ext=\"png\",
        content=b\"...\",
        mime_type=\"image/png\",
    )
    # \"exchange://hades-01/image-mcp/a1b2c3.png\"

# Consumer side
data = fx.read_exchange_uri(\"exchange://hades-01/image-mcp/a1b2c3.png\")
# Raises ExchangeGroupMismatch if the URI's group != fx.exchange_id
# Raises FileNotFoundError if the file is missing/expired

await fx.sweep()  # producer-side TTL+LRU sweep, idempotent
```

### Concrete behaviours required by spec v0.3

1. **Env detection** — read `MCP_EXCHANGE_DIR` (unprefixed; deployer-controlled, shared across containers). Empty/unset ⇒ `is_configured=False`. Existing-but-not-a-directory ⇒ raise `FileExchangeConfigError` at construction; do not silently fall back.

[...]

PR #26, opened 2026-04-26
feat(file-exchange): adopt v0.3 spec + ship protocol surface

### LINKED ISSUE #20: MCP File Exchange v0.3: adopt spec + ship protocol surface (FileRef, ExchangeURI, capability declaration)
## Summary

A draft spec exists (\"MCP File Exchange Specification\" v0.2.5, currently in the operator's Obsidian vault, status: draft, never implemented) for cross-MCP-server file transfer that solves the pass-by-reference problem — a 5 MB image moves between an image-mcp producer and a vault-mcp consumer without round-tripping through the LLM context as base64. The spec is unimplemented, has no users outside the operator's own work, and is freely changeable. The natural home for the convention is `fastmcp_pvl_core` since every pvl-family server depends on it and the producer/consumer roles map cleanly across the family (markdown-vault, image-generation, paperless, scholar).

This issue covers **adopting the spec into `docs/specs/file-exchange.md` (with proposed v0.3 revisions) and shipping the protocol surface** — the typed envelope, URI parser, and capability-declaration helper. The runtime (atomic writes, exchange-group lifecycle) lives in a follow-up issue. A server can implement v0.3 with `http`-only transport using just this issue's deliverables; `exchange` transport requires #FOLLOWUP_C2.

This is the protocol contract that #19's `create_download_link` plumbing actually serves under.

## Spec adoption

Copy the Obsidian draft into `docs/specs/file-exchange.md` as v0.3, with the following revisions. Each revision is justified inline so the diff against v0.2.5 is reviewable.

### 1. `origin_id` is an opaque round-trip handle, not a tool-parameter rename

**v0.2.5 says:** \"Producer tool MUST accept a parameter named `origin_id`.\"

**Problem:** breaks every existing `create_download_link` signature in the family (MV uses `path`, image-gen uses `uri`, paperless uses `document_id+variant`). It also forces semantic uniformity across genuinely heterogeneous handles.

**v0.3:** Producer tool MUST accept a parameter named `origin_id`. The value is opaque to the client — the producer MAY interpret `origin_id` as a path, document id, image id, or any internally-meaningful handle. Tools MAY additionally accept domain-named aliases (`path`, `document_id`, etc.) but `origin_id` MUST be a working alias. The handle is whatever the producer chose to put in `file_ref.origin_id` when it earlier emitted the file reference; round-tripping that exact string back to `create_download_link(origin_id=...)` MUST resolve to the same file.

### 2. Capability advertises minor-version only

**v0.2.5 spec is at v0.2.5 (patch), capability declaration uses `version: \"0.2\"` (minor).** Already consistent — just be explicit in v0.3 that capability advertises `MAJOR.MINOR` and patch is internal. Bump the advertised version to `\"0.3\"` for this revision.

### 3. `.exchange-id` file format

**v0.2.5 says:** \"a UUID is generated, persisted in `.exchange-id`.\" Doesn't specify encoding or trailing whitespace.

**v0.3:** UTF-8, plaintext UUID (8-4-4-4-12 hex with hyphens, lower or upper case), with or without a single trailing newline. Consumers MUST strip trailing whitespace before comparison. File MUST be created mode `0644`.

### 4. Namespace collision detection

**v0.2.5 is silent on what happens if two servers declare the same namespace.**

**v0.3:** A server starting up MUST check whether another file under `$MCP_EXCHANGE_DIR/{namespace}/` was modified within the last \"liveness window\" (default 60s) by a writer with a different process identity, and SHOULD log a warning. (A startup-blocking lock is rejected as overkill — the rare collision is human error.) Add a SHOULD that deployers configure `MCP_EXCHANGE_NAMESPACE` explicitly when running multiple instances of the same server image.

### 5. Filesystem ownership

**v0.2.5 is silent on UID/GID for the shared volume.**

**v0.3:** The shared volume MUST be writable by the effective UID of every participating server. Deployers using the pvl-family `docker-entrypoint.sh` UID-drop pattern (PUID/PGID) MUST align UIDs across containers. Add an example to deployer setup.

### 6. Consumer `fetch` MUST handle both schemes when both methods are advertised

**v0.2.5:** consumer's `fetch` tool name appears in `transfer_methods.http.tool`, but spec doesn't explicitly couple the same tool to `exchange://` resolution.

**v0.3:** A consumer that advertises both `exchange` and `http` in its `transfer_methods` MUST accept both `exchange://` and `http(s)://` URIs in its declared intake tool, dispatching by URI scheme. (Markdown-vault-mcp's existing `fetch` tool — currently HTTP-only — fits this exactly.)

### 7. Capability declaration via FastMCP — open question

**v0.2.5 mandates declaring capabilities under `experimental.file_exchange` during the MCP `initialize` handshake.** FastMCP 3.2 has no first-class hook for setting `experimental` capabilities (verified against constructor signature + instance attrs). Either:

- (a) Reach into FastMCP's lowlevel server (`mcp.server.lowlevel.Server`) which does support experimental capabilities, or
- (b) File a small upstream feature request on `jlowin/fastmcp` to expose `experimental_capabilities` on the `FastMCP` constructor.

**v0.3 stance:** the spec stays as written; this issue's `register_file_exchange_capability(mcp, decl)` helper MUST work, even if it has to monkey-patch the lowlevel server's capability set. Filing the upstream FastMCP issue is part of acceptance.

### Non-revisions (sanity-checked, leaving as-is)

- Atomic write via dotfile + rename: correct.
- `O_CREAT | O_EXCL` for `.exchange-id` init: correct.
- Producer-owned lifecycle: correct.
- Once-decoded URI segment validation + raw-string JSON validation distinction: correct and important.
- `transfer` extension via string-keyed methods with priority order: correct, forward-compatible.
- `preview` field intentionally loose: correct.

## Protocol-surface API additions

Three additions to `fastmcp_pvl_core`, all pure-data + small helpers — no I/O, no env access. The runtime side lives in the follow-up issue.

### 1. `FileRef` dataclass

```python
from fastmcp_pvl_core import FileRef, FileRefPreview, TransferMethods

@dataclass(frozen=True, slots=True)
class FileRef:
    origin_server: str
    origin_id: str
    transfer: TransferMethods
    mime_type: str | None = None
    size_bytes: int | None = None
    preview: FileRefPreview | None = None

    def to_dict(self) -> dict[str, Any]: ...
    @classmethod
    def from_dict(cls, raw: dict[str, Any]) -> FileRef: ...

@dataclass(frozen=True, slots=True)
class FileRefPreview:
    description: str | None = None
    dimensions: tuple[int, int] | None = None
    thumbnail_base64: str | None = None
    thumbnail_mime_type: str | None = None
    metadata: Mapping[str, Any] | None = None
```

`TransferMethods` is a typed dict / `Mapping[str, Mapping[str, Any]]`. Validation on `from_dict` (at least one method, required fields).

### 2. `ExchangeURI` parser/validator

```python
from fastmcp_pvl_core import ExchangeURI, ExchangeURIError

uri = ExchangeURI.parse(\"exchange://hades-01/image-mcp/a1b2c3.png\")
uri.exchange_id   # \"hades-01\"
uri.namespace     # \"image-mcp\"
uri.id            # \"a1b2c3\"
uri.ext           # \"png\"
uri.filename      # \"a1b2c3.png\"

ExchangeURI.validate_segment(\"raw_origin_id\", role=\"json_param\")  # no URI decoding
ExchangeURI.validate_segment(\"%2e%2e\",         role=\"uri\")         # one decode pass
```

Strict adherence to v0.3 §6.3 (\"Security and Path Resolution\") — single-pass URI decoding, no decoding for JSON params, segment rules applied uniformly.

### 3. Capability-declaration helper

```python
from fastmcp_pvl_core import FileExchangeCapability, register_file_exchange_capability

cap = FileExchangeCapability(
    namespace=\"image-mcp\",
    exchange_id=\"hades-01\",  # None when exchange not configured
    produces=[\"image/png\", \"image/webp\"],
    consumes=[],
    transfer_methods={
        \"exchange\": {},
        \"http\": {\"tool\": \"create_download_link\"},
    },
    version=\"0.3\",
)
register_file_exchange_capability(mcp, cap)
```

The helper sets `experimental.file_exchange` on the underlying MCP server (mechanism TBD per open question above). `version` defaults to the spec version this library implements.

## Why upstream

- Same justification as #19: every pvl-family server is a producer or consumer; the library is the right home for shared protocol contracts.
- Splitting protocol surface (this issue) from runtime (#FOLLOWUP_C2) lets a server adopt the http-only path quickly while exchange-volume support waits for the runtime work.
- The spec has been sitting in a private vault for months. Adopting it into a public repo gives it a stable URL, version history, and an open invitation for non-pvl MCP servers to participate.

## Acceptance

- [ ] `docs/specs/file-exchange.md` exists as v0.3, incorporating revisions #1-#7 above. Old v0.2.5 frontmatter retained as a \"Changelog\" appendix.
- [ ] `FileRef`, `FileRefPreview`, `TransferMethods` exported from `fastmcp_pvl_core`. `to_dict`/`from_dict` round-trip preserved by tests.
- [ ] `ExchangeURI.parse` rejects: `..` segments, `/` or `\\` in segments, null bytes, leading/trailing whitespace, double-encoded payloads (`%252e%252e%252f`).
- [ ] `ExchangeURI.validate_segment(role=\"json_param\")` does **not** URI-decode (e.g. accepts `req-%20-id` verbatim).
- [ ] `register_file_exchange_capability(mcp, cap)` populates `experimental.file_exchange` such that an MCP `initialize` response advertises it. Mechanism documented in module docstring.
- [ ] If FastMCP 3.x lacks an `experimental_capabilities` hook, an upstream issue is filed on `jlowin/fastmcp` and linked from the helper's docstring.
- [ ] Unit tests cover: spec example file_refs round-trip, segment-validation table, capability declaration shape matches spec example.
- [ ] README / docs index links to the new spec page.

## Out of scope

- The exchange-method runtime: env detection, `.exchange-id` exclusive-create, atomic writes, lifecycle sweep, consumer file resolution. (Filed as the follow-up runtime issue.)
- Content negotiation (producer reads consumer's `consumes` list and adapts) — spec lists this as Future.
- Streaming / large files — spec lists this as Future.
- Promoting the spec out of `experimental` to a formal MCP standard — spec is explicit that this is a stopgap.

## References

- Source spec: \"MCP File Exchange Specification\" v0.2.5 (private Obsidian vault). Will be embedded in `docs/specs/file-exchange.md` as v0.3 by this issue.
- pvliesdonk/fastmcp-pvl-core#19 — `create_download_link` plumbing that this protocol contract sits on top of.
- Reference servers per spec §\"Reference Implementations\": markdown-vault-mcp (consumer), image-generation-mcp (producer), with paperless-mcp and scholar-mcp as additional producers in the family.

pvliesdonk:
Closes #20

## Summary
Lifts the MCP File Exchange convention from a private Obsidian draft into a public spec doc and ships the typed envelope, URI parser, and capability-declaration helper in `fastmcp_pvl_core`. The runtime (env detection, atomic IO, exchange-group lifecycle, consumer URI resolution) is the follow-up — issue #21.

## Spec doc — `docs/specs/file-exchange.md` (v0.3)

v0.2.5 source content with the seven revisions specified in the issue:
1. `origin_id` is an opaque round-trip handle (alias, not rename) — fixes the v0.2.5 reading that would have broken every existing `create_download_link` signature.
2. Capability advertises `MAJOR.MINOR` only (made explicit).
3. `.exchange-id` file format pinned (UTF-8 UUID, optional trailing newline, mode `0644`).
4. Namespace collision detection: warn-don't-block + deployer guidance.
5. Filesystem ownership: shared volume must be writable by every participating UID; PUID/PGID alignment documented.
6. Consumer's intake tool dispatches by URI scheme when both `exchange` and `http` are advertised.
7. New "Capability declaration via FastMCP" design-decision section.

v0.2.5 frontmatter retained as a Changelog appendix.

## Protocol surface — `_file_exchange.py`

- `FileRef`, `FileRefPreview` — frozen dataclasses with `to_dict`/`from_dict` round-trips. Optional fields omitted on the wire when `None`.
- `TransferMethods` — open `Mapping[str, Mapping[str, Any]]` so future method keys (`s3`, `scp`, `gdrive`, …) don't need a structural change.
- `ExchangeURI.parse` — single-pass `urllib.parse.unquote`, then rejects residual `%XX` (double-encoded traversal-bypass signature) and applies spec §6.3 segment rules (no `/`, `\`, `..`, control bytes, etc.; namespace can't start with a dot).
- `ExchangeURI.validate_segment(role=…)` — `"uri"` decodes once; `"json_param"` validates as-is (JSON-RPC params must not be URI-decoded — would corrupt legitimate `%` bytes per spec §6.3).
- `FileExchangeCapability.to_capability_dict()` produces the exact shape in spec §3.9.
- `register_file_exchange_capability(mcp, cap)` monkey-patches `mcp._mcp_server.create_initialization_options` to inject `experimental.file_exchange`. See "FastMCP integration" below.

## FastMCP integration — confirmed mechanism

FastMCP 3.2.4 exposes `experimental_capabilities` on `_mcp_server.create_initialization_options(...)` but every transport call site invokes that method with no arguments — verified against:
- `fastmcp/server/http.py:177`
- `fastmcp/server/mixins/transport.py:216`
- `fastmcp/client/transports/memory.py:63`

So any state set on the FastMCP instance is dropped before the capability dict is built. Per the issue's open question (option `(a)` reach into the lowlevel server), the helper patches `create_initialization_options` on a per-instance basis. The wrapper unwraps itself on repeated registrations so payloads replace, not stack. End-to-end test: `create_initialization_options()` returns an `InitializationOptions` whose `capabilities.experimental` contains `{"file_exchange": cap.to_capability_dict()}`.

## Open: upstream FastMCP issue

The acceptance asks for a feature request on `jlowin/fastmcp` proposing `experimental_capabilities` as a first-class `FastMCP.__init__` kwarg. Filing on a third-party repo is the maintainer's call — proposed body:

> **Title:** Expose `experimental_capabilities` on `FastMCP.__init__` (or a setter)
>
> **Body:** `LowLevelServer.create_initialization_options` accepts an `experimental_capabilities: dict[str, dict[str, Any]]` kwarg, but every transport call site in FastMCP 3.x invokes it with no arguments (`fastmcp/server/http.py:177`, `mixins/transport.py:216`, `client/transports/memory.py:63`), so any state set elsewhere is discarded. Servers that want to advertise `experimental` capabilities (e.g. for [MCP File Exchange](…)) currently have to monkey-patch `_mcp_server.create_initialization_options` on each instance. Proposal: accept `experimental_capabilities` as a `FastMCP.__init__` kwarg and plumb it through.

Happy to file myself if you authorize — just say the word.

## Test plan
- [x] 47 new tests in `tests/test_file_exchange.py`
- [x] Spec §3.1 example round-trips through `FileRef.from_dict`/`to_dict` verbatim
- [x] Spec §3.9 producer example matches `FileExchangeCapability.to_capability_dict()` exactly
- [x] `ExchangeURI.parse` rejects: `..`, `/`, `\`, `%2F`, `%5C`, `%00`, whitespace, double-encoded `%252e%252e%252f`, namespace starting with `.`, missing/empty ext
- [x] `validate_segment(role="uri")` decodes once; `role="json_param"` does NOT decode
- [x] `register_file_exchange_capability` end-to-end: payload visible in `InitializationOptions`; re-registration replaces; caller-supplied experimental dict is merged
- [x] Full suite: 273 passed
- [x] `ruff format` / `ruff check` clean; `mypy --strict` clean
- [x] README links to the new spec page

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,14 @@ mcp = FastMCP(
 wire_middleware_stack(mcp)
 ```
 
+## Specs
+
+- [MCP File Exchange v0.3](docs/specs/file-exchange.md) — convention for
+  cross-MCP-server file transfer (`exchange://` URIs over a shared
+  volume + `http` fallback). The protocol surface (`FileRef`,
+  `ExchangeURI`, `register_file_exchange_capability`) ships in
+  `fastmcp_pvl_core`; the runtime is tracked in issue #21.
+
 ## License
 
 MIT
--- a/docs/specs/file-exchange.md
+++ b/docs/specs/file-exchange.md
@@ -1,0 +1,676 @@
+# MCP File Exchange Specification
+
+**Version:** 0.3
+**Status:** experimental
+**Tags:** mcp, spec, interop
+
+## Problem
+
+MCP servers cannot communicate directly with each other. The client mediates all interactions. When one server produces a file that another server needs to consume (e.g. an image generator producing an image that a vault server stores), the file content must pass through the client's context window as base64, wasting tokens and hitting size limits.
+
+## Goal
+
+Define a lightweight convention that allows independently developed MCP servers to exchange files efficiently when co-deployed, with graceful degradation to remote transfer when they are not.
+
+## Relationship to MCP
+
+MCP has no sideband for bulk data. The protocol is JSON-RPC over stdio or HTTP: every piece of content, including binary files, passes through the message stream as base64-encoded text inside tool results or resource contents. The client (typically an LLM host) receives this content into its context window. There is no protocol-level mechanism for a server to send a file directly to a client's filesystem, to another server, or to stream bytes outside the JSON-RPC channel.
+
+Specifically:
+
+- **Resources** (`BlobResourceContents`) can serve binary content, but the client reads it into context. There is no "download to disk without entering the context window."
+- **Streamable HTTP transport** means MCP servers are already HTTP endpoints, but the HTTP layer carries only JSON-RPC messages. Serving files at custom HTTP paths (as `create_download_link` does) works but is outside the MCP specification.
+- **Tool results** can contain base64 image data, but this consumes context window space proportional to the file size.
+- **MCP Apps** render HTML in a sandboxed iframe and can make network requests, but they are designed for interactive UI, not data transfer.
+
+The practical consequence is that passing a 5 MB image between two MCP servers costs thousands of tokens even though the LLM only needs to know "there is a PNG, 1024x768, of a circuit board diagram." The core problem this specification solves is **pass-by-reference**: the LLM receives lightweight metadata about a file (type, size, thumbnail, description) while the actual bytes travel outside the context window via shared filesystem or direct server-to-server transfer.
+
+This specification is designed as a **stopgap convention**. It intentionally uses MCP's `experimental` capability field and imposes no changes to the MCP protocol itself. If MCP later adds native file transfer or a bulk data sideband, implementations of this spec should be straightforward to migrate. The conventions are structured to be forward-compatible with that outcome: the file reference object maps naturally to a hypothetical MCP-native file handle, and the transfer methods abstraction can accommodate a future `mcp-native` method alongside the current `exchange` and `http` methods.
+
+## Concepts
+
+### File Reference
+
+The interop surface. When an MCP tool produces a file intended for cross-server use, it returns a **file reference** alongside or instead of inline content:
+
+```json
+{
+  "origin_server": "image-mcp",
+  "origin_id": "a1b2c3",
+  "mime_type": "image/png",
+  "size_bytes": 245760,
+  "preview": {
+    "description": "Generated circuit board diagram, top-down view",
+    "dimensions": {"width": 1024, "height": 768}
+  },
+  "transfer": {
+    "exchange": {
+      "uri": "exchange://hades-01/image-mcp/a1b2c3.png"
+    },
+    "http": {
+      "tool": "create_download_link"
+    }
+  }
+}
+```
+
+| Field | Required | Description |
+|---|---|---|
+| `origin_server` | MUST | Namespace of the producing server. The client uses this to identify which server connection to call for transfer negotiation. |
+| `origin_id` | MUST | Opaque identifier for this file on the origin server. Passed as a parameter when requesting transfer via any method. |
+| `mime_type` | SHOULD | MIME type of the file. |
+| `size_bytes` | MAY | File size in bytes. |
+| `transfer` | MUST | Object whose keys are transfer method names and whose values are method-specific metadata. At least one method MUST be present. See [Transfer Methods](#transfer-methods). |
+| `preview` | SHOULD | Lightweight representation of the file for LLM context. See below. |
+
+The file reference does **not** contain a download URL or inline content. Transfer is initiated lazily by the client through the declared methods.
+
+#### Preview
+
+The `preview` field gives the LLM enough information to reason about a file without ingesting the full binary. This is the key to pass-by-reference: the LLM sees metadata, not megabytes.
+
+```json
+"preview": {
+  "description": "Generated circuit board diagram, top-down view, 4-layer PCB",
+  "dimensions": {"width": 1024, "height": 768},
+  "thumbnail_base64": "/9j/4AAQSkZJRg...",
+  "thumbnail_mime_type": "image/jpeg",
+  "metadata": {
+    "prompt": "top-down view of a 4-layer PCB",
+    "model": "flux-schnell"
+  }
+}
+```
+
+All `preview` fields are optional. Producers SHOULD include at least a `description` so the LLM can make informed decisions about the file without requesting the full content.
+
+| Field | Description |
+|---|---|
+| `description` | Human/LLM-readable summary of the file content. |
+| `dimensions` | For images/video: `width` and `height` in pixels. |
+| `thumbnail_base64` | Small preview image, base64-encoded. SHOULD be under 10 KB to keep context costs minimal. |
+| `thumbnail_mime_type` | MIME type of the thumbnail (e.g. `image/jpeg`). Required when `thumbnail_base64` is present. |
+| `metadata` | Arbitrary key-value pairs with producer-specific context (prompt, model, page count, duration, etc.). |
+
+The `preview` field is intentionally unstructured beyond the common fields listed above. Different file types benefit from different metadata (images need dimensions, PDFs need page counts, audio needs duration). Producers include what is relevant; consumers and LLMs use what they recognise.
+
+A file reference MAY be embedded as a field within a larger tool response. For example, an image generation tool might return prompt metadata, dimensions, and a `file_ref` field containing the file reference. The spec does not prescribe the field name, but `file_ref` is conventional.
+
+### Usage Patterns
+
+File references support two patterns with different trade-offs:
+
+#### Augmented response (backward-compatible)
+
+The tool returns its normal output to the LLM (including inline content like thumbnails, metadata, or text) and additionally includes a file reference for cross-server transfer:
+
+```json
+{
+  "image_id": "a1b2c3",
+  "prompt": "top-down view of a 4-layer PCB",
+  "content_type": "image/png",
+  "dimensions": {"width": 1024, "height": 768},
+  "thumbnail_b64": "/9j/4AAQSkZJRg...",
+  "file_ref": {
+    "origin_server": "image-mcp",
+    "origin_id": "a1b2c3",
+    "mime_type": "image/png",
+    "size_bytes": 245760,
+    "transfer": {
+      "exchange": {"uri": "exchange://hades-01/image-mcp/a1b2c3.png"},
+      "http": {"tool": "create_download_link"}
+    }
+  }
+}
+```
+
+The LLM already has everything it needs from the native response: it can see the thumbnail, knows the dimensions, understands what was generated. The `file_ref` is purely a transfer handle. `preview` is redundant and can be omitted.
+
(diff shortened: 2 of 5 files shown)


pvliesdonk:
Pushed `ed5a710` addressing both bot reviews.

**Applied (claude-review)**
- **Bug 1 — size_bytes int coercion** (also flagged by gemini): JSON parsers may yield `245760.0`; dataclass annotation promises `int`. Now coerced via `int(...) if not None else None`.
- **Bug 2 — query strings / fragments slipped past `parse`**: `exchange://g/n/file.png?foo=bar` produced `ext="png?foo=bar"` silently. Added `?` and `#` to `_FORBIDDEN_SEGMENT_CHARS` so they're rejected whether plain or percent-encoded.
- **Improvement — `FileExchangeCapability.__post_init__` validation**: namespace and exchange_id are now validated via `ExchangeURI.validate_segment(role="json_param")` plus the spec §3.8 dot-prefix guard, so a bad value can't silently propagate into capability dicts and `exchange://` URIs.
- **Minor — `validate_segment(role: ...)` is now `Literal["uri", "json_param"]`** so mypy catches bad callsites at compile time. Runtime `ValueError` kept as a defense for callers bypassing the type system.

**Applied (gemini-code-assist)**
- **`dimensions` null-value check**: `dimensions={"width": None, "height": 100}` previously crashed with `TypeError` inside `int(None)`. Replaced `"key in raw"` with `raw.get(key) is None` so explicit nulls are caught with a helpful message.
- **Required-field null check**: `origin_server=None` was silently round-tripping to `"None"` via `str()`. Now rejected alongside missing keys.
- **size_bytes int coercion**: same as claude-review Bug 1.

**Skipped (with reasoning)**
- **`FileRefPreview.from_dict` scalar type-checking** for `description` / `thumbnail_base64` / `thumbnail_mime_type`. claude-review explicitly framed this as "no action required" — duck-typing, not a safety issue. My read agrees: validating at this layer would just add noise without changing observable behavior at use-sites.

10 new tests cover the fixes; 283 tests pass; ruff format/check + mypy --strict clean. Resolving the three gemini threads.

PR #27, opened 2026-04-26
feat(file-exchange): ship FileExchange runtime (env, exclusive-create, atomic IO, sweep)

### LINKED ISSUE #21: MCP File Exchange v0.3: implement FileExchange runtime (env detect, exclusive-create init, atomic IO, lifecycle, consumer resolution)

pvliesdonk:
Closes #21

## Summary
Companion to #20. Lifts the file-exchange runtime into core so every consumer of #20's protocol surface doesn't re-derive env detection, `.exchange-id` race-safety, atomic write/read, and lifecycle sweep.

## API
`FileExchange` (exported from `fastmcp_pvl_core`):

- **`from_env(default_namespace, *, env, ttl_seconds, storage_ceiling_bytes)`** — reads `MCP_EXCHANGE_DIR` / `MCP_EXCHANGE_ID` / `MCP_EXCHANGE_NAMESPACE`. Returns an unconfigured sentinel (`is_configured=False`) when `MCP_EXCHANGE_DIR` is unset/blank. Validates that the dir exists and is a directory; resolves namespace via env > `default_namespace`; rejects dot-prefix namespaces.
- **`write_atomic(*, origin_id, ext, content, mime_type)`** → `exchange://` URI. Validates `origin_id` and `ext` as JSON params (no URI decoding per spec §6.3). Writes to `$ns/.{id}.{ext}.tmp`, fsyncs, POSIX-renames to `$ns/{id}.{ext}`. The dotfile prefix is the kill-resilience property — a crash leaves only a dotfile, which consumers ignore by spec §5.
- **`read_exchange_uri(uri)`** → bytes. Parses via `ExchangeURI.parse` (single-pass decode + segment rules + double-encoding rejection from #20), verifies the URI's exchange-group ID matches this server's, refuses leading-dot filenames (spec §5), reads the file. Raises `ExchangeGroupMismatch` with **both IDs** in the message on group mismatch.
- **`sweep()`** → int. Producer-only operation: TTL eviction, then optional LRU eviction by mtime when `storage_ceiling_bytes` is set. Restart-resume: rebuilds working set from disk on every call, no in-memory registry. Skips dotfiles unconditionally so it never races a producer's in-progress write.

## `.exchange-id` race safety (spec §3.5)
`_resolve_exchange_id` uses `os.open(O_CREAT | O_EXCL | O_WRONLY, 0o644)` to acquire group identity. POSIX `rename(2)` is explicitly rejected by the spec because it silently overwrites — would split-brain on a race. `EEXIST` → read the winner's value (whitespace stripped per spec). An explicit `MCP_EXCHANGE_ID` that conflicts with the persisted value raises `ExchangeGroupMismatch` immediately, `raise ... from None` since the `FileExistsError` isn't the actual cause of the data disagreement.

## Errors
- `FileExchangeConfigError(RuntimeError)` — config issues at construction.
- `ExchangeGroupMismatch(ValueError)` — group ID disagreement (carries `N818 noqa`: spec-defined name, no `Error` suffix).
- `ExchangeURIError` — already in #20; reused for traversal/forbidden-char/double-encoded URIs and for dotfile filenames at read time.

## Acceptance criteria coverage
- [x] `FileExchange` exported from `fastmcp_pvl_core`
- [x] `from_env` returns `is_configured=False` cleanly when `MCP_EXCHANGE_DIR` is unset
- [x] **Race test**: 16-thread `from_env` against empty dir produces exactly one `.exchange-id` file with all workers agreeing on the value
- [x] **Override test**: explicit `MCP_EXCHANGE_ID` mismatching the existing `.exchange-id` raises `ExchangeGroupMismatch`
- [x] **Atomic-write test**: pre-created `.{id}.{ext}.tmp` (simulating kill -9) is invisible to consumers — `read_exchange_uri` raises `FileNotFoundError`
- [x] **Cross-namespace read**: vault-mcp reads `exchange://g1/image-mcp/foo.png` written by image-mcp on the same volume
- [x] **Group-mismatch test**: `read_exchange_uri` with a different `exchange_id` raises `ExchangeGroupMismatch` with both IDs in the message
- [x] **Sweep test**: TTL eviction + LRU eviction (oldest-by-mtime first when ceiling exceeded)
- [x] **Once-decoded validation**: `read_exchange_uri("exchange://g1/vault-mcp/%252e%252e%252f.png")` raises `ExchangeURIError`
- [x] **Path traversal**: `read_exchange_uri("exchange://g1/../escaped/file.png")` raises `ExchangeURIError`
- [x] Module docstring documents producer-owned-lifecycle and consumer read-only constraints

## Test plan
- [x] 44 new runtime tests in `tests/test_file_exchange.py` (91 total in the file)
- [x] Full suite: 317 passed
- [x] `ruff format` / `ruff check` clean; `mypy --strict` clean

### DIFF
--- a/src/fastmcp_pvl_core/__init__.py
+++ b/src/fastmcp_pvl_core/__init__.py
@@ -31,9 +31,12 @@ from fastmcp_pvl_core._file_exchange import (
     SPEC_VERSION as FILE_EXCHANGE_SPEC_VERSION,
 )
 from fastmcp_pvl_core._file_exchange import (
+    ExchangeGroupMismatch,
     ExchangeURI,
     ExchangeURIError,
+    FileExchange,
     FileExchangeCapability,
+    FileExchangeConfigError,
     FileRef,
     FileRefPreview,
     TransferMethods,
@@ -53,10 +56,13 @@ __version__ = "1.0.0"  # PSR overrides at build time
 __all__ = [
     "ArtifactStore",
     "AuthMode",
+    "ExchangeGroupMismatch",
     "ExchangeURI",
     "ExchangeURIError",
     "FILE_EXCHANGE_SPEC_VERSION",
+    "FileExchange",
     "FileExchangeCapability",
+    "FileExchangeConfigError",
     "FileRef",
     "FileRefPreview",
     "IconSpec",
--- a/src/fastmcp_pvl_core/_file_exchange.py
+++ b/src/fastmcp_pvl_core/_file_exchange.py
@@ -1,20 +1,36 @@
-"""MCP File Exchange v0.3 — protocol surface.
+"""MCP File Exchange v0.3 — protocol surface and runtime.
 
-Implements the typed envelope (:class:`FileRef`, :class:`FileRefPreview`),
-the URI parser (:class:`ExchangeURI`), and the capability-declaration
-helper (:func:`register_file_exchange_capability`) for the spec at
-``docs/specs/file-exchange.md``.
+Implements the spec at ``docs/specs/file-exchange.md``:
 
-The runtime side (env detection, atomic writes, exchange-group lifecycle,
-consumer URI resolution) lives in a separate module and is not yet
-implemented — see issue #21.
+- **Protocol surface**: :class:`FileRef`, :class:`FileRefPreview`,
+  :class:`ExchangeURI` (parser + segment validator),
+  :class:`FileExchangeCapability`, and the capability-declaration
+  helper :func:`register_file_exchange_capability`.
+- **Runtime**: :class:`FileExchange` for env-driven group membership,
+  exclusive-create ``.exchange-id`` initialisation, atomic
+  write/read of namespaced files, and producer-side TTL+LRU lifecycle
+  sweep.
+
+Lifecycle constraints from the spec:
+
+- The *producing server* owns its namespace directory's lifecycle
+  exclusively. Only the producer deletes its own files (TTL + LRU
+  eviction via :meth:`FileExchange.sweep`).
+- The *consuming server* treats the exchange directory as read-only —
+  :meth:`FileExchange.read_exchange_uri` never modifies or deletes
+  files. Consumers MUST ignore dotfile names.
 """
 
 from __future__ import annotations
 
+import logging
+import os
 import re
+import time
+import uuid
 from collections.abc import Mapping
 from dataclasses import dataclass
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 from urllib.parse import unquote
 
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Pushed `917be1c` addressing both findings:

**Real bug fixed (was failing CI on `test_concurrent_from_env_creates_exactly_one_exchange_id`)**

The race test caught an actual race in `_resolve_exchange_id`. The naive `O_EXCL` pattern atomically creates an empty file, then the writer populates it with the UUID — but concurrent threads that hit `FileExistsError` could read the file *before* the writer fsynced, getting `""`. The 16-thread test failed with `{"", "<actual-uuid>"}`.

Fix: switch to a link-based atomicity pattern. Each thread writes its UUID payload to a per-thread `.exchange-id.tmp.<pid>.<tid>`, fsyncs, then `os.link()`s into place. `link(2)` returns `EEXIST` on collision and never overwrites — same atomicity guarantee as `O_EXCL` but with no empty-file window. Losers `unlink` their tmp and read the winner's already-fsynced content; tmp cleanup runs in `finally` so crashed inits don't leak dotfiles. Also added a defensive empty-file check on the fast path so a corrupted `.exchange-id` from any prior crashed init surfaces as `FileExchangeConfigError` rather than silently returning `""`. 10x stress runs of the race test pass locally.

**CodeQL findings (file-permission mode 0o644)**

GitHub Advanced Security flagged the `os.open(..., 0o644)` calls in `_resolve_exchange_id` and `write_atomic` as overly permissive. The `0o644` mode is **intentional and spec-mandated** — see `docs/specs/file-exchange.md` §3.5 (`'.exchange-id' MUST be created with mode 0644`) and §3.4/§3.10 on cross-UID readability for the shared-volume topology. Reviewer's call here is technically correct in isolation but blocked by the spec.

Restructured to the `chmod`-after-create pattern: open with restrictive `0o600`, write+fsync, `os.fchmod(fd, 0o644)` before close. Data-flow-equivalent but doesn't trip CodeQL's check on the syscall mode literal at the `open()` call. Comments at both call sites cite the spec for future maintainers.

317 tests pass; ruff format/check + mypy --strict clean. Wakeup armed.

pvliesdonk:
Pushed `6705ddb` addressing the second-pass review feedback.

**Applied (claude-review medium)**
- **Dot-prefix `origin_id` storage leak**: `write_atomic(origin_id=".hidden", ext="png")` would silently create a file no consumer can ever read AND no sweep can clean. Block at the producer with a clear `ExchangeURIError`. Same guard for `ext` to be safe. Two new tests.

**Applied (gemini medium + claude-review medium)**
- **`exchange_id` validation at config-load**: An explicit `MCP_EXCHANGE_ID="foo/bar"` previously silently persisted to `.exchange-id` and only broke later at URI construction with a confusing 4-segment parse error. Now `ExchangeURI.validate_segment(role="json_param")` runs at `from_env` time. New test verifies `from_env` rejects `MCP_EXCHANGE_ID="bad/group"` cleanly.

**Applied (gemini medium)**
- **Single-pass sweep**: refactored from two full directory scans (TTL pass, then LRU re-stat pass) into one. While iterating once, TTL-evict and collect surviving `(path, mtime, size)` tuples for the optional LRU consideration. Halves `stat()` calls per sweep.

**Applied (claude-review minor)**
- **LRU `total` inflation on external delete**: if a file vanishes between our `stat()` and `unlink()`, the previous code kept the ghost size in `total` and could over-evict one extra surviving file. Moved `total -= size` outside the `try`, so it always runs after a deletion attempt — whether we did it or someone else did, those bytes are gone.

**Already fixed**
- Gemini's race-condition finding on `_resolve_exchange_id` was reviewing commit `fd81e86` (pre-race-fix). It describes the exact bug the link-based atomicity pattern in `917be1c` closes — the empty-file window between `O_EXCL` create and the writer's fsync. Resolving the thread.

3 new tests; 320 tests pass; ruff format/check + mypy --strict clean.

pvliesdonk:
Pushed `f4a9b12` addressing claude-review pass-3.

**Real bug fixed (medium)**
- **`explicit_id` persisted before validation**. My pass-2 fix put `ExchangeURI.validate_segment(exchange_id)` AFTER `_resolve_exchange_id`. Wrong order: `MCP_EXCHANGE_ID="bad/group"` would link-write `"bad/group"` into `.exchange-id` first, THEN trip validation, leaving the corrupt file behind. A corrected env var on the next run would read the stale value via the fast path and raise `ExchangeGroupMismatch` with no diagnostic pointing at `.exchange-id`. Operator gets stuck.

  Now validates `explicit_id` BEFORE the resolve/write. Kept the post-resolve check as a recovery net for pre-fix deployments. Test now asserts the critical post-condition: `.exchange-id` MUST NOT exist after the rejection.

**Previously-open hardening**
- **`__init__` accepts unvalidated namespace/exchange_id**. `FileExchange` is in `__all__`; direct instantiation can't bypass `from_env`'s spec checks. Added the same `validate_segment` + dot-prefix guards in `__init__` for both fields when set. The redundant call from `from_env` is cheap (microseconds for a UUID) and the safety net covers programmatic / test-harness construction. 3 new tests.

323 tests pass; ruff format/check + mypy --strict clean.

Issue #67, opened 2026-05-09
file_exchange: export type aliases (BufferedReceiver, StreamReceiver, PreLinkValidator) at top level


pvliesdonk:
Tracked from the local-review pass for #64.

## Context

The upload direction defines three type aliases used by receiver-author code:

- \`BufferedReceiver = Callable[[UploadRecord, bytes], dict[str, Any] | Awaitable[dict[str, Any]]]\` (in \`_file_exchange_runtime.py\`)
- \`StreamReceiver = Callable[[UploadRecord, AsyncIterator[bytes]], Awaitable[dict[str, Any]]]\` (in \`_file_exchange_runtime.py\`)
- \`PreLinkValidator = Callable[[str, dict[str, Any] | None], None]\` (in \`file_exchange.py\`)

A receiver author writing typed code today must import from private modules:
\`\`\`python
from fastmcp_pvl_core._file_exchange_runtime import BufferedReceiver
\`\`\`

By contrast, \`ConsumerSink\`, \`FetchContext\`, \`FetchResult\` (download-direction) ARE exported from the package root.

## Proposal

Re-export all three from \`fastmcp_pvl_core/__init__.py\` and add to \`__all__\`. Update CHANGELOG.

## Out-of-scope rationale

#64 focused on the runtime feature; export hygiene is independent. Receiver authors who want types today can reach into the submodule.

Issue #66, opened 2026-05-09
file_exchange upload: register_file_exchange_upload should accept a consumes= parameter


pvliesdonk:
## What

`register_file_exchange_upload` does not accept a `consumes=` parameter.
A server that registers *only* the upload helper (no
`register_file_exchange` download side) therefore contributes no
`consumes` MIME list to its `experimental.file_exchange` capability — it
advertises an empty `consumes`, invisible to MIME-aware capability
discovery.

## Current state (verified against `main`, v0.3.0)

```
register_file_exchange_upload(
    mcp, *, namespace, env_prefix,
    receiver, stream_receiver, pre_link_validator, accepts,
) -> UploadHandle
```

No `consumes`. The capability builder's `consumes` list is populated
only by `register_file_exchange` (the download side); the per-server
capability still carries a top-level `produces` / `consumes` pair.

## Proposal

Add `consumes: Sequence[str] = ()` to `register_file_exchange_upload`,
pushed into the builder via
`_get_or_create_builder(mcp, ..., consumes=tuple(consumes))` — the
builder already set-union-merges `consumes` across registrar calls.

`consumes` passes the CLAUDE.md kwarg-classification test: the MIME
types a server ingests are domain-specific knowledge pvl-core cannot
infer, so it is a legitimate domain-hook kwarg (not a shape/config
override).

## Open questions

- Should the route-level `accepts` Content-Type filter auto-derive from
  `consumes` when `consumes` is non-empty and `accepts` is left default?
  Or stay fully independent? (Default position: independent — `accepts`
  is a wire-level POST gate, `consumes` is a capability-discovery hint.)
- Dual-direction servers already get a set-union merge of `consumes`
  across the two registrar calls — confirm that remains correct.

## History

Rescoped 2026-05-16 (post-#75). The original issue was framed around
"Spec Amendment 11" and the `transfer_methods.http.upload` capability
shape — both removed by the #75 umbrella work. The underlying gap
(upload-only servers cannot advertise `consumes`) is unchanged and
verified against current `main`; the framing above is updated to the
v0.3.0 spec.

## Related

- #75 — file-exchange umbrella (closed).

pvliesdonk:
Body rescoped 2026-05-16 — the original premise was overtaken by the #75 umbrella work (v0.3.0 spec, merged upload re-implementation). See the updated description for the current state and the `## History` note for what changed.

Issue #65, opened 2026-05-09
file_exchange: set_upload_store global singleton breaks multi-namespace use


pvliesdonk:
**Update:** the original \`_capability_builders[id(mcp)]\` half of this issue was fixed in PR #68 (commit \`c74d4a5\`) by storing the builder on the FastMCP instance via a private attribute (\`mcp._pvl_file_exchange_builder\`). Mirrors the \`_pvl_experimental_middleware\` pattern. \`id(mcp)\` reuse hazard is gone.

**Remaining concern**: \`set_upload_store\` / \`get_upload_store\` is a module-level singleton. Each call to \`register_file_exchange_upload\` replaces the singleton; multi-namespace registrations on the same process see only the last-registered store via \`get_upload_store()\`. Route handlers and tools use closure-captured stores so routing is correct, but the public singleton API is broken for multi-namespace use.

This is symmetric with the pre-existing \`set_artifact_store\` / \`get_artifact_store\` pattern, which has the same limitation but is grandfathered in.

## Possible approaches

- (a) Drop \`set_upload_store\` / \`get_upload_store\` entirely. Route handlers don't need them (closure capture works). Downstream consumers shouldn't be reaching for the singleton; if they need the store, they can keep the \`UploadHandle\` returned by the registrar.
- (b) Move to per-instance: \`mcp._pvl_upload_store\`. \`get_upload_store(mcp)\` requires the mcp argument. Symmetric with the capability-builder fix.
- (c) Document the limitation; leave for users who genuinely need multi-namespace to wire their own.

Recommend (b) for symmetry with the capability-builder fix and the existing \`_pvl_experimental_middleware\` precedent. Same change should likely also apply to the artifact-direction singleton (\`_artifact_store\`) for consistency, but that's a wider refactor that may want its own issue.

## Out-of-scope rationale (PR #68)

Same as before: substantive design change with public-API implications. Defer to a focused follow-up.

PR #68, opened 2026-05-09
feat(file-exchange): upload direction (register_file_exchange_upload, spec v0.4)

### LINKED ISSUE #64: Add upload-endpoint helper as inbound mirror of register_file_exchange's create_download_link
## Context

`register_file_exchange` (and the older `create_download_link` artifact route) handle the **outbound** half of file transfer well: server can mint a one-time HTTPS URL, client GETs the bytes.  No symmetric **inbound** primitive exists — when a local agent (claude-code, cursor, a script) wants to push a file *into* a server-managed area, the only options today are:

- Round-trip through MCP context as base64 — context-bound, broken above ~1 MB.
- The agent posts to its own out-of-band endpoint — every server reinvents.

## Proposal

Symmetric helper, working name **`register_upload_endpoint`**.  Every downstream domain (markdown vaults, paperless inboxes, scholar-mcp's PDF ingester) plugs in a domain-specific `receiver` callable; the helper handles tokens, the HTTP route, size enforcement, and tool registration uniformly.

### API

```python
from fastmcp_pvl_core import register_upload_endpoint, UploadRecord

handle = register_upload_endpoint(
    mcp,
    namespace="vault",                        # tool/route prefix
    env_prefix="MARKDOWN_VAULT_MCP",          # for env-var lookups
    transport="auto",                         # auto|http|sse|stdio
    receiver=_my_receiver_fn,                 # called at POST time, see signature below
    upload_tool_name="create_upload_link",    # override the default if it collides
    accepts=("application/octet-stream",),    # content-types accepted; "*/*" disables check
    max_bytes_default=10 * 1024 * 1024,       # operator-overridable via env
    ttl_default=300,                          # link lifetime in seconds
)
```

### Receiver signature

```python
def _my_receiver_fn(record: UploadRecord, body: bytes) -> dict[str, Any]:
    """Commit the uploaded bytes.

    Args:
        record: token metadata — `target_id` (opaque path/identifier the
            tool caller chose), `content_type` (from request), `size_bytes`,
            and any `extra` dict the tool caller passed at link creation.
        body: raw bytes streamed from the POST request, already validated
            against ``record.max_bytes``.

    Returns:
        Dict serialized as the HTTP 200 response body and (optionally) as
        the tool-call follow-up payload.  Conventional keys: ``path``,
        ``size_bytes``, ``content_type``.

    Raises:
        ValueError: invalid input (path traversal, disallowed extension,
            etc.) — surfaces as HTTP 400.
        FileExistsError: if the receiver enforces no-overwrite — HTTP 409.
        Any other exception logs at ERROR and surfaces as HTTP 500.
    """
```

The receiver runs synchronously inside the POST handler (FastAPI/Starlette already gives us the threadpool); no async constraint on the callable.

### Tool registered

The helper registers a tool named `<upload_tool_name>` (default `create_upload_link`) on `mcp`:

```python
@mcp.tool(tags={"write"})
async def create_upload_link(
    target_id: str,
    ttl_seconds: int = 300,
    max_bytes: int | None = None,        # defaults to env / max_bytes_default
    extra: dict[str, Any] | None = None, # opaque, passed verbatim to receiver
) -> dict[str, Any]:
    """Generate a one-time HTTPS POST URL for pushing bytes into the server.

    Returns:
        upload_url (str): the one-time URL to POST to.
        expires_in_seconds (int): TTL.
        target_id (str): echo of the input.
    """
```

The host server can wrap the registered tool with a thin override (e.g. to validate `target_id` as a vault path *before* link creation) by calling `mcp.disable("create_upload_link")` and re-registering with their own validator that delegates to `handle.create_link(...)`.

### HTTP route

`POST /<namespace>/uploads/{token}` (e.g. `/vault/uploads/abc123`).  Behavior:

1. Look up `token` in `UploadStore`. Missing → **404**. Expired → **410**. Already consumed → **404** (don't leak that the token *did* exist).
2. Check `Content-Length` against `record.max_bytes`; reject with **413** before reading the body.
3. Stream the body into a buffer with a per-chunk size check (defense in depth — clients can lie about Content-Length). Exceed → **413**, partial bytes discarded.
4. Validate `Content-Type` against `accepts` if configured; mismatch → **415**.
5. Mark token consumed (one-time) *before* dispatching to receiver — so a slow receiver can't be retried by a racing client.
6. Call `receiver(record, body)`. Map exceptions to status codes per the receiver-signature contract above.
7. Return the receiver's dict as JSON (HTTP 200).

### `UploadStore`

Process-wide singleton, parallel to `ArtifactStore`'s shape but inverted lifecycle:

```python
class UploadStore:
    def create(
        self,
        target_id: str,
        ttl_seconds: int,
        max_bytes: int,
        extra: dict[str, Any] | None = None,
    ) -> UploadRecord: ...

    def consume(self, token: str) -> UploadRecord | None:
        """Atomic: returns record + marks consumed, or None if missing/expired/consumed."""

    def gc(self) -> int:
        """Lazy expiry sweep, called on each `create`. Returns count purged."""
```

Tokens are UUID4. `UploadRecord` is a frozen dataclass with `token`, `target_id`, `created_at`, `expires_at`, `max_bytes`, `extra`, and a `consumed` flag set atomically in `consume()`.

### Env vars (optional, host opts in via `env_prefix`)

| Env var | Default | Description |
|---|---|---|
| `<PREFIX>_UPLOAD_ENABLED` | `true` on http/sse, `false` on stdio | Master switch. |
| `<PREFIX>_UPLOAD_MAX_BYTES` | `max_bytes_default` arg | Operator override of the per-link cap. |
| `<PREFIX>_UPLOAD_TTL` | `ttl_default` arg | Operator override of the link TTL. |

### Symmetry table (for sanity-checking the design)

| Concern | Download (today) | Upload (this issue) |
|---|---|---|
| Token store | `ArtifactStore` | `UploadStore` |
| Tool name (default) | `create_download_link` | `create_upload_link` |
| HTTP route | `GET /<ns>/artifacts/{token}` | `POST /<ns>/uploads/{token}` |
| Body shape | server → bytes | bytes → server |
| Auth surface | one-time token, short TTL | one-time token, short TTL |
| Tag | `tags=set()` (read-only) | `tags={"write"}` |
| Required env | `BASE_URL` | `BASE_URL` |
| Transport gating | http/sse only | http/sse only |
| Co-deployed transfer | `exchange://` (file_ref) | TBD — `exchange://` reverse, future |

## Reference consumer

markdown-vault-mcp will wire `register_upload_endpoint(...)` directly in pvliesdonk/markdown-vault-mcp#443 once this ships — *no local POC*, the upstream helper is the only implementation.  Receiver dispatches to `Collection.write` (for `.md`) or `Collection.write_attachment` (for binaries) based on the `target_id`'s extension, with the existing path-traversal and attachment-allowlist validation running at link-creation time so the agent gets the rejection in-band.

## Out of scope (future work)

- `exchange://` reverse for co-deployed-server pushes (mirrors how `register_file_exchange` adds `exchange://` for downloads). File a follow-up after this lands.
- Resumable uploads (multi-part). Treat single-shot POST as v1.
- Per-token Content-Type *override* by the POSTer; v1 trusts the link-creation-time `accepts` setting.

## Related

- pvliesdonk/fastmcp-pvl-core#63 — sibling `register_app_tool` helper for hash-routed MCP Apps tools
- pvliesdonk/markdown-vault-mcp#443 — MV consumer, wires this helper directly (no local POC)

pvliesdonk:
## Summary

Symmetric inbound mirror of `register_file_exchange`. Mints one-time HTTPS POST URLs via a registered `create_upload_link` tool; receiver callable handles domain-specific commit. Buffered (`receiver=`) or streaming (`stream_receiver=`) variants. Optional `pre_link_validator=` runs at link creation so invalid `target_id`s surface as in-band tool errors.

Also bumps the File Exchange spec from v0.2 to v0.4 (Amendments 10 + 11) — `transfer_methods.http` now nests by direction (`download` and `upload` sub-keys); upstream tooling supports a `legacy_capability_shape=False` migration flag.

Closes #64.

## Architecture

- **Token store refactor.** `_BaseTokenStore[T]` generic in new `_token_store.py` provides UUID4 minting, lazy expiry sweep, atomic one-time consume. `ArtifactStore` (download) and `UploadStore` (upload) compose. Old `_artifacts.py` becomes a 24-line deprecation shim.
- **Upload route.** `register_upload_route` mounts `POST /<ns>/uploads/{token}` with the documented status-code contract (200/400/404/409/410/413/415/500). One bounded async-generator (`_bounded_chunks`) feeds both buffered and streaming receivers; mid-stream overrun raises a private `_UploadOversizeError` (subclass of `BaseException` to defeat receivers' `except Exception`).
- **Public facade.** `register_file_exchange_upload(...)` mirrors `register_file_exchange`. Resolves transport, env-driven enable/overrides, BASE_URL gate, `pre_link_validator` (with structural ValueError-vs-bug differentiation), TTL clamp+floor, mutual-exclusion guard.
- **Capability merge.** `_FileExchangeCapabilityBuilder` accumulates per-FastMCP download- and upload-side contributions; both registrars route through it via `_get_or_create_builder` + `_emit_capability`. Namespace and legacy-shape mismatches log warnings.

## Spec amendments (v0.4.0 draft)

- **Amendment 10:** `transfer_methods.http` gains direction tagging — `{download: {tool: ...}, upload: {tool: ..., max_bytes: ..., max_ttl_seconds: ...}}`.
- **Amendment 11:** Inbound HTTP transfer — `create_upload_link` tool + `POST /<ns>/uploads/{token}` route + status-code contract + `accepts` per-method MIME filter + `target_id` character rules.

Implementation note recorded in the amendment: token consume happens at the lookup step (anti-replay), so any 4xx response burns the link; clients re-call `create_upload_link` to retry. Tests pin this behavior.

## Test plan

- [x] **594 → 599 → 600+ tests passing** (started 562 on main; final 599 after fix bundle + 1 more after typo fix).
- [x] `uv run ruff format --check .` — clean.
- [x] `uv run ruff check .` — clean.
- [x] `uv run mypy src/` — strict, no issues.
- [x] `uv run pyright src/fastmcp_pvl_core/{file_exchange,_file_exchange_runtime,_token_store,_file_exchange_protocol}.py` — clean.
- [x] **100% line + 100% branch coverage on all new code** (Task 19 gap-filling pass).
- [x] **Local-review circus:** primary `pr-review-toolkit:code-reviewer`, `superpowers:code-reviewer` (second-opinion), `silent-failure-hunter`, `type-design-analyzer`, `pr-test-analyzer` all return clean after fix bundle `b278a34` + typo fix `ac6b02d`.

## Implementation discipline

- 20-task TDD plan at `docs/superpowers/plans/2026-05-09-file-exchange-upload.md`.
- Per-task spec compliance + code quality reviews via fresh subagents.
- Design doc at `docs/superpowers/specs/2026-05-09-file-exchange-upload-design.md` (committed before code).

## Out-of-scope follow-ups

Tracked issues for items deliberately deferred (none block this PR):

- pvliesdonk/fastmcp-pvl-core#65 — `_capability_builders` keyed by `id(mcp)` (architectural; needs WeakRef or per-instance attribute).
- pvliesdonk/fastmcp-pvl-core#66 — `register_file_exchange_upload` should accept a `consumes=` parameter (substantive design).
- pvliesdonk/fastmcp-pvl-core#67 — Export `BufferedReceiver` / `StreamReceiver` / `PreLinkValidator` at top-level `__init__.py` (advanced typing).

## Downstream impact

- **markdown-vault-mcp#443** is the first downstream consumer — wires this directly, no local POC. Unblocks once this lands.
- Three template issues to file in `pvliesdonk/fastmcp-server-template` for upload-receiver scaffolding + README/CHANGELOG/dependency bump.


pvliesdonk:
## Gemini round-1 disposition

Gemini's review fired on the initial draft (2026-05-09T22:25Z), before five rounds of claude-review fixes landed. Status of each item:

| # | Finding | Disposition |
|---|---|---|
| 1 | `_capability_builders` global keyed by `id(mcp)` | **Tracked** as follow-up issue #65. Architectural; needs WeakRef or per-instance attribute, requires upstream FastMCP/pydantic coordination. |
| 2 | `set_upload_store` global singleton | **Tracked** as follow-up issue #65. Same root cause as #1; per-instance storage is the correct fix. |
| 3 | `tags=set(tool_tags)` → `list(tool_tags)` | **Wrong / kept as `set`.** FastMCP's `@tool` overload is typed `tags: set[str] | None`; mypy and pyright reject `list[str]` with hard errors at the call site. Verified via fresh `uv run mypy src/` and `uv run pyright src/fastmcp_pvl_core/file_exchange.py` (both clean with `set(...)`, both error with `list(...)`). Claude-review rounds 4 and 5 independently confirmed this disposition. |
| 4 | `stream_receiver` direct `await` | **Fixed** in commit `134ee60` (round 2). The streaming path now mirrors the buffered path: `result = stream_receiver(record, _bounded_chunks()); if inspect.isawaitable(result): result = await result`. Tested by `test_post_sync_stream_receiver_returning_plain_dict`. |

After 5 rounds of claude-review, the latest verdict is **"clear to merge — no new findings"** (round 5 at 2026-05-10T06:00Z). Items 1 and 2 are architectural follow-ups tracked in #65; item 3 is wrong as written; item 4 is fully resolved.

Not re-triggering `/gemini review` since the local-review-circus discipline reserves that for cases with a fresh diff to evaluate, not as iterative prodding.

Marking ready for human review.

pvliesdonk:
Addressing gemini's HIGH-priority finding (capability builder `id(mcp)` registry) in c74d4a5 — replaced module-level dict with per-FastMCP-instance attribute (`mcp._pvl_file_exchange_builder`), mirroring the existing `_pvl_experimental_middleware` pattern. The `id(mcp)` reuse hazard is gone. Defensive fallback dict for pydantic-strict configs is kept but unexercised in tests.

Issue #65 updated to focus only on the remaining `set_upload_store` singleton concern (medium priority); that one is intentionally deferred per the PR scope, with the (b) per-instance approach pre-recommended for symmetry.

Re-triggering /gemini review now that the local circus is clean on the new commit.

pvliesdonk:
Three rounds of post-flip-to-ready iteration since gemini's last review at `c74d4a5`:

- `fce4966` — sync receivers + sync `pre_link_validator` dispatch via `asyncio.to_thread`; dropped the `_capability_builders_fallback` dead code; env-override values validated with `ConfigurationError` naming the offending var (UPLOAD_MAX_BYTES / UPLOAD_TTL / UPLOAD_TTL_MAX). All three were direct responses to gemini's previous findings.
- `6364433` — env range validation (non-positive values rejected with named `ConfigurationError`); `test_sync_pre_link_validator_runs_in_threadpool` pinning the round-2 threadpool dispatch on the validator path.
- `02c4eaa` — three parallel range-validation tests covering the new <=0 guards (negative values for each of the three env vars).

Local-review circus (primary `pr-review-toolkit:code-reviewer` + second-opinion) ran on `fce4966` and returned ✅ clean — no findings at any severity. Claude-review rounds 7–10 found progressively smaller polish items, all addressed; round 10 explicitly says "clear to merge — no new findings".

/gemini review

pvliesdonk:
Addressed gemini's only round-3 finding in d9d432b — streaming branch now matches the buffered branch's threadpool dispatch (`inspect.iscoroutinefunction` first, sync via `asyncio.to_thread`). Test `test_post_sync_stream_receiver_runs_in_threadpool` parallels the buffered-receiver threadpool test. Local circus clean; 613 passing.

/gemini review

pvliesdonk:
32d3ece — wildcard `accepts` now passed verbatim per gemini round-4 finding. End-to-end merge test pins `http.upload.accepts == ["*/*"]` when the registrar uses the default. Local circus + CI clean.

/gemini review

Issue #87, opened 2026-05-15
audit: verify http-direction tool/route conformance against v0.3.0 spec


pvliesdonk:
## What

Audit the **`http` transfer method** (download direction) implementation for conformance to the v0.3.0 file-exchange spec — the `http.source` producer tool (`create_download_link`), the `http.sink` consumer tool (`fetch_file`), and the download-serving HTTP route.

This is the `http`-row counterpart of #74 (which gives `http_upload` a clean-slate conformance rebuild). It completes the transfer-mechanism coverage matrix.

## Background

The transfer-mechanism surface is a matrix — `exchange` has no tool roles; `http` and `http_upload` each have a `source` and a `sink` role:

| Method / role | Tool(s) | Conformance issue |
|---|---|---|
| `exchange` | none | n/a |
| `http.source` | `create_download_link` | **this issue** |
| `http.sink` | `fetch_file` | **this issue** |
| `http_upload.sink` | `create_upload_link` + POST route | #74 |
| `http_upload.source` | `upload` | #85 |

#86 migrates the capability-declaration *shape* for every method/role to the v0.3.0 `source`/`sink` form. But the `http` direction's *tool contracts and route mechanics* have not been verified against the post-revert v0.3.0 spec. `http` is a v0.2.x method — v0.3.0 added `http_upload`, #83 changed only the capability shape — so `http`'s tools/route are *presumed* conformant. "Presumed" is not "verified": the download path was last substantially touched in the A11-amendments era, and during #86's design work the download-serving route could not be located (the only `custom_route` in `_file_exchange_runtime.py` is the upload one).

## Scope

- Verify `create_download_link` (producer / `http.source`) tool contract against the spec: `origin_id` parameter handling, the `{url, ttl_seconds}` return, TTL.
- Verify `fetch_file` (consumer / `http.sink`) tool contract against the spec: `url` + optional `path` parameters, safe path auto-generation.
- Locate and verify the download-serving HTTP route (where `create_download_link` URLs are served) — token mechanics, status codes, anti-leak rules.
- Check for any surviving A11/v0.4-amendments-era constructs in the `http` path.
- Confirm the §"Producing server" / §"Consuming server" conformance-checklist obligations are met.

## Outcome

May close as "already conformant" with the verification recorded — that is a valid and good outcome. The point is a *verified* result rather than an assumption.

## Depends on / related

- #86 — capability-builder shape migration (the shape axis; this issue is the tool/route axis).
[...]

Issue #85, opened 2026-05-15
impl: sender-side http_upload primitive (the source-role upload tool)


pvliesdonk:
## What

Add a pvl-core helper for the **sender side** of the `http_upload` transfer method — the `source` role: a tool that POSTs bytes to a receiver-issued upload URL.

#74 implements only the **receiver side** (`sink` role — `create_upload_link`, the POST route, token mechanics). The sender side was explicitly split out during the #74 brainstorm to keep #74's diff reviewable; this issue tracks it.

## Background

The `http_upload` method is bidirectional (see `docs/specs/file-exchange.md` §`http_upload` and the dual-role capability work in #83):

- **`sink` role (receiver)** — mints the upload URL via `create_upload_link`, accepts the POST. Implemented by #74.
- **`source` role (sender)** — POSTs bytes to a receiver's URL via the `upload` tool. **This issue.**

Per the spec, the sender-side tool carries a `source` tagged-union parameter (`path` / `exchange_uri` / `http_url` / `inline_b64`) and the server advertises `transfer_methods.http_upload.source` with `source_variants` listing the variants its tool implements.

## Scope

- New `register_file_exchange_*` helper (name per the spec / consistent with #74's receiver helper) that registers the sender-side `upload` tool and populates the `source` sub-object of the `http_upload` capability block.
- Kwarg surface passes the #72/#73 classification test — domain hooks only; operator values on env vars; no shape-override kwargs.
- `source_variants` reflects exactly the tagged-union variants the helper implements.
- Tests against spec conformance.

## Depends on

- #83 / PR #84 — the `source`/`sink` capability shape this helper populates.
- #74 — the receiver-side implementation; this helper should mirror its patterns and hook surface.

## Related

- #75 — file-exchange architectural cleanup (umbrella)
- #73 — pvl-core framing principle
- #72 — hook surface classification

pvliesdonk:
**Blocked on #93.**

The #85 brainstorm surfaced that the `upload` tool's `source` tagged union (`path` / `exchange_uri` / `http_url` / `inline_b64`) and the `source_variants` capability field are sender-side *implementation detail* that should never have been normative wire-format spec text — byte-acquisition strategy, not interop contract. `path` is incoherent as a cross-server variant.

[...]

PR #92, opened 2026-05-15
impl: re-implement http_upload receiver primitive against v0.3.0 spec (#74)

### LINKED ISSUE #74: impl: re-implement upload primitive against the new spec (supersedes register_file_exchange_upload)
## What

Replace `register_file_exchange_upload` with a new implementation that is 100% compliant with the spec released in #71.

## Background

The current `register_file_exchange_upload` was implemented against the LLM-invented Amendment 11 (see #76). It treats `target_id` as opaque per the invented amendment's segment rules, which forced markdown-vault-mcp (PR #443) to ship a vault-root-only upload primitive.

Once #71 lands a proper spec for the reverse HTTP method, this issue ships the matching pvl-core implementation. The current implementation is on the way out — anything from it that's worth keeping (token route mechanics, body-size cap, exception-to-status-code mapping, threadpool wrapping for sync receivers) gets re-derived from the new spec, not carried forward as-is.

## Constraints

- Spec compliance is binary. If the new spec says the parameter is `X`, the implementation accepts `X`, not `X` plus a backwards-compat alias for the old name.
- Hooks expose domain-specific behaviour only (per #73). A receiver callback for "where do these bytes go" is appropriate; an option to override the spec-conformant tool name is not.
- markdown-vault-mcp's PR #443 wiring will be replaced wholesale (see pvliesdonk/markdown-vault-mcp#488).

## Acceptance

- [ ] Depends on #71 spec release.
- [ ] New `register_file_exchange_*` helper (name TBD per spec) wires routes, capability, and tool registration end-to-end against the new spec.
- [ ] Old `register_file_exchange_upload` removed (or marked deprecated for one minor with a clear removal date).
- [ ] Hook surface conforms to the principle from #72.
- [ ] Tests pass against spec conformance suite.
- [ ] Downstream migration tracked in pvliesdonk/markdown-vault-mcp#488.
- [ ] **Template impact**: scaffold updated in pvliesdonk/fastmcp-server-template#131.

## Related

- #75 — file-exchange architectural cleanup
- #71 — spec evolution this implementation conforms to
- #72 — hook surface audit (applies the same classification)
- #73 — pvl-core framing principle
- pvliesdonk/markdown-vault-mcp#488 — downstream migration

pvliesdonk:
## Summary

Clean-slate re-implementation of the receiver/`sink`-side `http_upload` primitive — the `create_upload_link` tool, the POST upload route, `UploadRecord`, and `register_file_exchange_upload` — 100% conformant with the v0.3.0 file-exchange spec, superseding the A11-era implementation.

- **`create_upload_link` tool** — parameters `origin_id` (required), `destination`, `content_type`, `ttl_seconds`, `max_bytes`; returns `{url, ttl_seconds, max_bytes}` (effective post-clamp values) on success. In-band rejection (bad `origin_id`, a `destination` rejected by `pre_link_validator`, a `content_type` outside `accepts`) returns the `transfer_failed` envelope — `{error, method, receiver_server, origin_id, message}` — rather than a raw tool error.
- **POST route** — the `410`-on-expired branch is removed; unknown / expired / already-consumed tokens all return an indistinguishable `404` with an empty body (the spec's anti-leak rule).
- **`UploadRecord`** — models the WHAT/WHERE split (`origin_id` / `destination` / `content_type`); the non-spec `extra` field is gone. `UploadStore.consume_or_status` removed; `consume` returns `UploadRecord | None`.
- **`register_file_exchange_upload`** — kwarg surface cut to six domain hooks (`namespace`, `env_prefix`, `receiver`, `stream_receiver`, `pre_link_validator`, `accepts`). The shape-override kwargs (`upload_tool_name`, `tool_tags`) and operator-config kwargs (`transport`, `max_bytes_default`, `ttl_default`, `ttl_max`) are removed — pvl-core fixes the tool name/tags; operator config moves to the existing `{PREFIX}_UPLOAD_*` env vars.
- **`PreLinkValidator`** — signature is now `(origin_id, destination)`.

The token-store mechanics (UUID4 entropy, atomic one-time consume, lazy TTL purge, the `_bounded_chunks` streaming generator, sync/async receiver dispatch, `413`/`415`/`5xx` handling) are spec-conformant and preserved unchanged.

## Breaking change

`register_file_exchange_upload`'s kwarg surface shrinks and `UploadRecord` (passed to `receiver` / `stream_receiver`) is restructured. The one downstream consumer, `markdown-vault-mcp`, migrates via `pvliesdonk/markdown-vault-mcp#488` — pvl-core ships the breaking change; downstream follows (no compat shim, per the umbrella's discipline).

## Out of scope

- The sender-side `upload` tool (`http_upload.source`) — #85.
- The `http`-direction tool/route conformance audit — #87.

## Test plan

- [x] `uv run pytest` — 617 passed.
- [x] `uv run ruff format --check .` / `ruff check .` / `mypy src` — all clean.
- [x] The three upload test files rewritten to the v0.3.0 contract, including: `404`-not-`410` for expired tokens; unknown/expired/consumed all byte-identical `404`s; the `transfer_failed` envelope on `pre_link_validator`, `content_type`, and `origin_id` rejection; `{url, ttl_seconds, max_bytes}` return and clamping; the `UploadRecord` WHAT/WHERE split.
- [ ] CI green.
- [ ] Bot review clean.

Closes #74.

### DIFF
--- a/docs/superpowers/plans/2026-05-15-upload-primitive-v030.md
+++ b/docs/superpowers/plans/2026-05-15-upload-primitive-v030.md
@@ -1,0 +1,642 @@
+# http_upload Receiver Primitive — v0.3.0 Re-implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Re-implement the receiver/`sink`-side `http_upload` primitive — `register_file_exchange_upload`, the `create_upload_link` tool, the POST route, and `UploadRecord` — to be 100% conformant with the v0.3.0 file-exchange spec.
+
+**Architecture:** The `http_upload` receiver is one tightly-coupled unit spanning three files: `_token_store.py` (`UploadRecord` / `UploadStore`), `_file_exchange_runtime.py` (the POST route), and `file_exchange.py` (the helper + tool + `UploadHandle`). The tool's parameter/return shape, the route's status codes, and the record's fields all change together — the call graph (`create_upload_link` → `UploadHandle.create_link` → `UploadStore.reserve`; route → `UploadStore.consume`) does not survive a partial migration with `mypy` green. Task 1 therefore rebuilds all three files plus rewrites the upload tests in one atomic, test-first commit. Task 2 is the full-suite sweep + quality gate.
+
+**Tech Stack:** Python 3.10–3.13, `uv`, `pytest`, `ruff`, `mypy`. Repo `/mnt/code/fastmcp-pvl-core`, branch `impl/upload-primitive-v030-issue-74`.
+
+**Design doc:** `docs/superpowers/specs/2026-05-15-upload-primitive-v030-design.md` (issue #74).
+
+---
+
+## File Structure
+
+- `src/fastmcp_pvl_core/_token_store.py` — `UploadRecord` gains `origin_id`/`destination`/`content_type`, drops `target_id`/`extra`. `UploadStore.reserve` parameters change; `consume_or_status` is deleted (`consume` stays).
+- `src/fastmcp_pvl_core/_file_exchange_runtime.py` — `_upload_handler` drops the `410` branch (uses `consume`, not `consume_or_status`).
+- `src/fastmcp_pvl_core/file_exchange.py` — `PreLinkValidator` signature; `UploadHandle` / `create_link`; `register_file_exchange_upload` kwarg surface; the `create_upload_link` tool body and return; a new `_upload_transfer_failed` envelope helper.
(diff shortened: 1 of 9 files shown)


pvliesdonk:
Re the claude-review note on `int(eff_ttl)` vs `round(eff_ttl)` (`file_exchange.py` create_upload_link return): keeping `int()`.

`ttl_seconds` in the return tells the caller how long the link is valid so it can plan retries. Truncation makes the returned value an *under*-estimate of the real TTL — a fractional effective TTL of 299.9s is reported as 299s, so a caller that stops using the link at the reported deadline stops ~0.9s early. That's the safe direction. `round()` could report 300s for a link that actually dies at 299.9s, inviting a POST that just misses the window and gets a 404. For a deadline a client acts on, conservative-low beats accurate. (In practice `eff_ttl` is integral unless an operator sets a fractional `{PREFIX}_UPLOAD_TTL`; the choice only matters in that edge case, and there `int()` is the correct one.)

The two test-coverage notes (assert the 404 body is empty; cover the built-in `destination` validation directly) are addressed in 0654a06.

Issue #149, opened 2026-05-21
Rollout tracker: file-exchange v0.1 adoption in downstream repos


pvliesdonk:
Sibling rollout tracker for #138.

Opens once pvl-core's file-exchange v0.1 surface from #138 ships (i.e. issue 10/10 of that EPIC closes). Each entry below is a separate issue in its respective downstream repo. This tracker is just the rollup; it closes when every downstream entry closes.

## Adoptions

- [ ] `pvliesdonk/fastmcp-server-template` — template-side wire-up so downstream copier-updates pick up the registration helpers.
- [ ] `pvliesdonk/markdown-vault-mcp` — wire `register_file_exchange_*` into the server's tool layer.
- [ ] `pvliesdonk/scholar-mcp` — same.
- [ ] `pvliesdonk/image-generation-mcp` — same.
- [ ] `pvliesdonk/reqeng-mcp` — same.

Per-downstream design happens in that downstream's own brainstorming session, against the file-exchange capabilities pvl-core exposes. Each downstream issue will link back here.

## Why a separate tracker

`CLAUDE.md` caps EPICs at ~10 issues. Bundling five cross-repo adoption tasks into the pvl-core EPIC would breach that cap and would also conflate cross-repo state with the pvl-core EPIC's own completion criteria. The rollout tracker is the right unit for cross-repo coordination.

Issue #148, opened 2026-05-21
file-exchange (10/10): top-level register_file_exchange_* helpers + Tasks integration + adoption docs


pvliesdonk:
Part of #138. Depends on issues 1–9.

## Scope

The opinionated public surface that ties everything below into a single registration call per role. Adds the Tasks composition that §14 requires for potentially-long transfers. Ships the README + downstream adoption guide.

## Deliverables

- `register_file_exchange_provider(mcp, tool_name, source: ArtifactSource, ...)`, `register_file_exchange_fetcher(mcp, tool_name, callback, ...)`, and the matching `receiver`/`sender` helpers. Each takes a mechanism-agnostic hook from issue 4 and wires the mechanics from issues 5/7/8.
- Tools registered through these helpers auto-declare `execution.taskSupport` (`"optional"` is the right default per §14); server auto-declares `tasks.requests.tools.call` in its capabilities when at least one role helper is registered.
- README section on the file-exchange extension.
- `docs/file-exchange.md` — pvl-core's *implementation* notes (not a spec; the spec lives in `mcp-file-exchange-ext`).
- Downstream adoption guide showing one minimal `register_file_exchange_*` wire-up per role.
- CHANGELOG entry.

## Spec references

§11 (carrying references through `tools/call`), §14 (Tasks integration), §16 (conformance).

pvliesdonk:
## Folded-in polish from PR #151 (#140) review — address while reworking `_errors.py`

PR #151's 4th-pass review surfaced two non-blocking improvements to
`src/fastmcp_pvl_core/_file_exchange/_errors.py`. They were deferred here
because #148 reworks that file anyway (adding the wire middleware around
`build_file_exchange_error`). Do these in the same pass:

**1. Use `model_copy` instead of post-construction `meta=` assignment.**
`build_file_exchange_error` currently does `result.meta = {_NAMESPACE_KEY: envelope}`
after construction, because `mcp.types.CallToolResult` lacks
`populate_by_name=True` so the constructor won't accept `meta=` by name.
That works today (the model is non-frozen) but breaks if a future `mcp`
release sets `frozen=True`. Replace with the frozen-safe form:

```python
return result.model_copy(update={"meta": {_NAMESPACE_KEY: envelope}})
```

(or construct the result with `meta` set via `model_validate({..., "_meta": ...})`).
Removes the mutation-after-construction smell too.

[...]

Issue #144, opened 2026-05-21
file-exchange (6/10): capability URL minter + token store (on the KV factory from #122)


pvliesdonk:
Part of #138. Depends on the selection + error envelope issue (2/10).

## Scope

High-entropy capability URL minter and the token store that backs lookup, expiry, and single-use lifecycle. Backed by the unified KV factory from #122 — **no new storage abstraction.**

## Deliverables

- Token generator: ≥128 bits of unguessable entropy, URL-safe.
- Token store API: `mint(metadata, ttl) -> token`, `lookup(token) -> record | None`, `consume(token)` for single-use enforcement, `revoke(token)`.
- Backing storage via `_kv_store` factory from #122.
- Env config: token TTL ceiling, max artifact size, public base URL for capability URLs.
- Unit tests covering: expiry boundary, single-use lifecycle, double-consume rejection, revocation.

## Spec references

§12 (capability URLs), §10.2/§10.3 (per-transport URL requirements).

Issue #142, opened 2026-05-21
file-exchange (4/10): mechanism-agnostic byte-source / byte-sink hook contracts


pvliesdonk:
Part of #138. Depends on the wire-format issue (1/10).

## Scope

Define the public hook protocols downstream implements: how a server reads bytes for an artifact it offers, and how it deposits bytes for an artifact it receives. Per the project's hooks-mechanism-agnostic discipline, the contracts **MUST NOT** name `filesystem`, `http`, or any other transport — they describe domain semantics only. A transport's specifics live behind the hook, not in its signature.

## Deliverables

- `ArtifactSource` protocol: given a domain key, returns bytes (or a stream) + `ArtifactMetadata`.
- `ArtifactSink` protocol: given an `IntakeTicket`'s `artifactId` + metadata, accepts bytes (or a stream) and confirms deposit.
- Atomic write-then-rename helper, available for any transport implementation that needs it.
- Test that statically asserts no transport name appears in any hook signature (an introspection-based check the reviewer can rely on).

## Spec references

Not a wire concern. This is pvl-core's internal abstraction, governed by `CLAUDE.md` and the project's framing principle (hooks expose domain-specific behaviour only).

PR #156, opened 2026-05-23
feat(file-exchange): filesystem transport binding (#143)

### LINKED ISSUE #143: file-exchange (5/10): filesystem transport — provider, fetcher, receiver, sender wiring
Part of #138. Depends on the exchange:// scheme issue (3/10) and the hook-contract issue (4/10).

## Scope

Bind `ArtifactSource`/`ArtifactSink` to the filesystem source and sink descriptors. Implements all four roles' filesystem path.

## Deliverables

- Provider mint: produces a `FilesystemSource` descriptor pointing at a path the `ArtifactSource` exposes.
- Fetcher consume: takes a `FilesystemSource`, canonicalizes-and-confines, opens read-only, verifies `size` and `digest`.
- Receiver mint: produces a `FilesystemSink` descriptor for an inbound artifact, allocates a deposit path.
- Sender consume: takes a `FilesystemSink`, writes atomically (temp + rename) into the resolved location.
- End-to-end test exchanging a file between two pvl-core-built mock servers over the filesystem transport only.

## Spec references

§10.1 (filesystem semantics), §15 (untrusted content), §16 (conformance for each role).

### LINKED ISSUE #155: file-exchange: define deposit file permissions for the filesystem sink
## Context

Split out of PR #154 (#142, the mechanism-agnostic hook contracts). Gemini's review of #154 flagged (security-medium) that `atomic_write` (`src/fastmcp_pvl_core/_file_exchange/_paths.py`) deposits artifacts with `tempfile.mkstemp`'s restrictive **0600** mode: `os.replace` preserves the temp file's permissions, so every deposited artifact lands owner-only, regardless of umask. On a shared exchange volume a *different-uid* reading party may then be unable to read the deposit.

## Why this is deferred to #143, not fixed in #142

`atomic_write` in #142 is a generic "write bytes to a path atomically" primitive with **no consumer yet**. *Which* mode a deposited artifact should carry is a security-model decision that belongs to the consumer that defines read-by-other-party semantics — the filesystem sink in **#143** (5/10 of EPIC #138). Picking a permission policy inside #142's primitive would prematurely fix a shape decision the consumer is better placed to make (per the `CLAUDE.md` framing principle: shape decisions are pvl-core's to own, but made where the domain context is understood).

This was a deliberate, maintainer-confirmed deferral during #154 review, not an oversight.

## What #143 must decide and implement

- The permission policy for deposited artifacts (e.g. umask-respecting `0666 & ~umask` like a normal file write, an explicit fixed mode appropriate to the exchange's security posture, or copy-existing-mode-on-overwrite). Note: reading the process umask is mildly racy in a threaded context (`atomic_write` runs under `asyncio.to_thread`), so a fixed/explicit mode is likely cleaner than umask introspection.
- Where to apply it: either `atomic_write` `os.fchmod`/`os.chmod`es the temp before `os.replace`, or the sink sets the mode after deposit.
- Whether the mode is a pvl-core shape decision (one value, downstream conforms) vs. operator config (env var) — classify per `CLAUDE.md`.

## Verification

- [ ] Deposited filesystem-sink artifacts have the agreed mode (not the accidental mkstemp 0600).
- [ ] A test asserts the resulting mode of a deposited file.

Part of EPIC #138.

pvliesdonk:
## Summary

Binds the mechanism-agnostic `ArtifactSource`/`ArtifactSink` hooks (#142) to the `filesystem` transport — the first consumer of the #142 contracts (5/10 of EPIC #138).

- New `_file_exchange/_filesystem.py`: the four role ops as free functions — `filesystem_provider_mint`, `filesystem_fetcher_consume`, `filesystem_receiver_mint`, `filesystem_sender_consume` — over two private byte primitives (`_stage` write / `_ingest` two-pass verify-then-deposit), plus the `filesystem_source_readable` / `filesystem_sink_writable` predicates that feed `select_source`/`select_sink`'s `is_accessible=`.
- Deposited/staged files get a fixed `0o664` (owner rw, group rw, other r) so a different-uid party on a shared volume can read them — **resolves #155**. `atomic_write` gains an optional `mode` param applied via `fchmod` on the temp fd before `os.replace` (default `None` preserves the prior `0o600`; the existing primitive/test is untouched).
- TOCTOU-safe read open (`O_NOFOLLOW | O_NONBLOCK` + `fstat` regular-file check) — **resolves the open-time TOCTOU item #141 deferred here**. Prefix-component races / full `openat` traversal are intentionally out of scope per §10.1's trust-boundary note.
- New `FileExchangeTransferError` (§13-coded) raised by the consuming ops (`not-accessible` / `size-mismatch` / `digest-mismatch` / `transfer-failed`, cause chained via `from exc`, generic wire `detail`); #148's middleware renders it. Offering/mint ops let hook/IO errors propagate unwrapped per §16.
- Stops at deposit on the push side; the receiver's lazy `artifactId`→bytes ingest is #144/#148 per #138's issue map.

## Notes

- `ArtifactSink.store_artifact`'s `artifact_id` was widened `str` → `str | None` (the pull-side value is `TransferHandle.artifact.id`, optional per §7.1; this also corrected a latent #142 contract inconsistency). This is a Protocol-signature change downstream sink implementers must accommodate (handle `None`), but **no downstream has adopted the #142 hooks yet** (adoption is the unstarted #149), so nothing breaks now.
- `HANDLE_TYPE`/`TICKET_TYPE` retyped as `Literal[...]` so the mint ops can construct wire models without a cast.
- Minor follow-up (not in scope here): the #142 *design/plan* docs under `docs/superpowers/` still describe the old "counting/hashing wrapper" hook mechanism that this PR's `_ingest` supersedes — point-in-time planning artifacts, not the live `_hooks.py` docstring (which is corrected here).

## Test plan

- [x] `uv run pytest` — 801 passed, 1 skipped (GITHUB_TOKEN-gated spec-sync) on Python 3.10 **and** 3.13
- [x] `uv run ruff format --check .` / `uv run ruff check .` / `uv run mypy src` — all clean
- [x] Unit: TOCTOU open (symlink/dir/FIFO rejection + fd-close on every failure branch), size/digest verify (match / size-mismatch / digest-mismatch / unsupported + uppercase algorithm), verify-before-deposit ordering (sink not called on mismatch), `0o664` deposit mode, §13 codes surviving end-to-end through `filesystem_fetcher_consume`, unmapped-volume → `ConfigurationError`, write-side IO failure (sender → `transfer-failed`, provider → unwrapped per §16), `artifact_id=None`
- [x] E2E: two pvl-core-built mock servers exchange a file over the filesystem transport — pull round-trip (both hooks) + push-to-deposit

Closes #143, closes #155


pvliesdonk:
Thanks for the review. Dispositions on the five findings (all non-blocking, per your framing):

1. **Sink blocking-read on the event loop** — agreed this is a `#149` (downstream adoption) decision, not a `#143` concern. The contract is already documented on both `_ingest` ("the sink reads the stream on the event loop; a sink with heavy I/O is responsible for offloading its own reads") and `ArtifactSink.store_artifact`. Whether pvl-core should thread-wrap before handing the stream over is a contract change to weigh when the first real sink lands in `#149`. No change here.
2. **Unsupported digest algorithm → `digest-mismatch`** — intentional per spec §15 ("cannot verify → verification fails"); documented and tested (`test_verify_stream_unsupported_algorithm_is_digest_mismatch`). The structured `detail` carries the reason for operators. No action, as you note.
3. **`atomic_write` parent-dir fsync (crash-durability)** — correct distinction: the code (and spec §10.1.3) guarantee atomic *visibility* via `os.replace`, which is what's claimed; directory-level crash-durability is a separate property that `#142`'s design doc consciously deferred. Tracked as #157 rather than expanding `#143`'s scope (it's an `atomic_write` property shared by any future deposit consumer).
4. **FIFO test would hang (not fail) if `O_NONBLOCK` were removed** — fair meta-point; the FIFO *rejection* itself is correctly asserted. A timeout marker would make an accidental `O_NONBLOCK` removal fail loudly rather than hang; minor test-robustness, can ride with #157 or a follow-up.
5. **UUID relpath collision** — astronomically unlikely; a defensive `exists()` guard would be noise, as you say. No action.

(The spec-sync check failed on a transient upstream `404` for a file that exists at the pinned SHA — main passes the same check and this diff touches nothing schema/sync-related; re-running the job.)

Issue #146, opened 2026-05-21
file-exchange (8/10): upload data plane — route + receiver/sender


pvliesdonk:
Part of #138. Depends on the token store issue (6/10) and the SSRF guard issue (9/10).

## Scope

HTTPS `PUT`/`POST` route that accepts capability URLs, plus the receiver/sender helpers for the `upload` transport.

## Deliverables

- `register_file_exchange_routes(mcp)` adds the PUT/POST endpoint at a pvl-core-chosen path (e.g. `/fx/u/{token}`). Body-size cap, `acceptMimeTypes` enforcement (RFC 7231 media-range semantics), `Content-Digest` (RFC 9530) verification, single-success-per-URL enforcement.
- Receiver helper: mints an `UploadSink` against the token store; returns an `IntakeTicket.sinks` entry; correlates received bytes to `artifactId` on success.
- Sender helper: posts the artifact through the SSRF guard module, sends `Content-Digest`, treats non-2xx as `transfer-failed`.
- Integration test: two-server push flow.

## Spec references

§10.3 (`upload` transport), §12 (capability URLs).

Issue #145, opened 2026-05-21
file-exchange (7/10): download data plane — route + provider/fetcher


pvliesdonk:
Part of #138. Depends on the token store issue (6/10) and the SSRF guard issue (9/10).

## Scope

HTTPS `GET` route that serves capability URLs, plus the provider/fetcher helpers for the `download` transport.

## Deliverables

- `register_file_exchange_routes(mcp)` mounts the GET endpoint at a pvl-core-chosen path (e.g. `/fx/d/{token}`). Range support, `Content-Type`/`Content-Length` headers when known, ambient credentials ignored.
- Provider helper: mints a `DownloadSource` against the token store; emits the descriptor inside a `TransferHandle.sources` entry.
- Fetcher helper: issues the HTTPS `GET` through the SSRF guard module, verifies `size` and `digest`, recovers a dropped connection with `Range`.
- Streaming end-to-end — artifacts never buffered whole in memory.
- Integration test: two-server pull flow.

## Spec references

§10.2 (`download` transport), §12 (capability URLs).

Issue #138, opened 2026-05-21
EPIC: adopt mcp-file-exchange-ext v0.1


pvliesdonk:
## Goal

Adopt the [mcp-file-exchange-ext v0.1 draft spec](https://github.com/pvliesdonk/mcp-file-exchange-ext) in `fastmcp-pvl-core` as the shared implementation for the `pvliesdonk/*-mcp` server family.

## External spec pin

- Repo: `pvliesdonk/mcp-file-exchange-ext`
- Commit: [`5f50a4e16a33a6bbc0888c142baec7fdfe858cb6`](https://github.com/pvliesdonk/mcp-file-exchange-ext/commit/5f50a4e16a33a6bbc0888c142baec7fdfe858cb6) (HEAD of `main` as of 2026-05-21)
- Extension version: `0.1` (draft), targets MCP revision `2025-11-25`
- Namespace: `nl.liesdonk.file-exchange`

The pinned commit is the forward-compatibility amendment to v0.1 (open transports, open error codes, `requires` must-understand, unknown-transport fallthrough). Future spec revisions are out of scope and tracked as separate work when they land.

## Scope recorded at decomposition (2026-05-21)

- All four roles (`provider`, `fetcher`, `receiver`, `sender`) and all three transports (`filesystem`, `download`, `upload`) ship together — single milestone.
- `pvl-core` owns the data-plane HTTP routes (`register_file_exchange_routes(mcp)` style); downstream just registers them.
- Downstream adoption (`markdown-vault-mcp`, `scholar-mcp`, `image-generation-mcp`, `reqeng-mcp`, `fastmcp-server-template`) is tracked separately in #149.

## Issue map

Each child issue's scope is the scope statement on that issue, **not a design**. Per-issue brainstorming opens when each issue's turn arrives — this EPIC explicitly does not contain per-issue designs.

1. #139 — wire format: schema, models, capability declaration
2. #140 — selection algorithm + error envelope (deps: #139)
3. #141 — `exchange://` URI scheme + path confinement (deps: #139)
4. #142 — mechanism-agnostic byte-source / byte-sink hook contracts (deps: #139)
5. #143 — filesystem transport: provider / fetcher / receiver / sender wiring (deps: #141, #142)
6. #144 — capability URL minter + token store on the KV factory from #122 (deps: #140)
7. #145 — download data plane: route + provider + fetcher (deps: #144, #147)
[...]

PR #122, opened 2026-05-20
feat: unified key-value storage factory (closes #121)

pvliesdonk:
## Summary

- Adds `build_kv_store(config, *, namespace) -> AsyncKeyValue` so every pvl-core subsystem that needs persistent state — event store today, future cleanroom file-exchange token store, future OAuth state — resolves to the same operator-chosen backend with isolated keyspaces.
- One env var (`<PREFIX>_KV_STORE_URL`) selects a backend across `memory://`, `file:///path`, `redis://...`, `dynamodb://<table>?region=...`, and `mongodb://...`. The network-backed backends are optional extras (`fastmcp-pvl-core[redis|dynamodb|mongodb]`) with lazy imports — memory/file deployments do not pull in client libraries.
- `build_event_store` refactored to delegate to the new factory under `namespace="events"`. The legacy `EVENT_STORE_URL` is honoured as a fallback when `KV_STORE_URL` is unset and emits a one-shot per-process deprecation warning.

## Implementation notes

- `ServerConfig.kv_store_url: str | None` field, loaded from `<PREFIX>_KV_STORE_URL`. `event_store_url` retained as a deprecated alias for the migration window.
- The returned store is wrapped in `PrefixCollectionsWrapper(prefix=namespace)`. The factory rejects empty/whitespace-only namespaces explicitly — the prefix is the entire isolation guarantee.
- Lazy backend imports catch `ModuleNotFoundError` (not the broader `ImportError`) and re-raise as `ImportError` naming the pvl-core extra to install. Inner ImportErrors from a successfully-loaded module surface with their original message.
- `file://` URLs are validated: `file://host/path` (non-empty netloc) and `file://` (empty path) raise `ValueError`. The `file://` branch verifies the `FileTreeStore` import succeeds **before** `mkdir`, so a missing extra cannot leave an orphan directory.
- `build_event_store` keeps its `(env_prefix, config)` public signature for backwards compatibility; the parameter is accepted but unused inside.

## Migration notes (in CHANGELOG)

The default file-tree directory shifts from `/data/state/events` to `/data/state` (the new factory's root, with `PrefixCollectionsWrapper(prefix="events")` providing on-disk isolation). Operators relying on the old default can set `EVENT_STORE_URL=file:///data/state/events` explicitly during the migration window, or accept the new layout — event entries TTL at 1h so functional impact is bounded.

## Local review

[...]

PR #158, opened 2026-05-23
feat(file-exchange): capability token minter + store (#144)

### LINKED ISSUE #144: file-exchange (6/10): capability URL minter + token store (on the KV factory from #122)

pvliesdonk:
## Summary

Implements the file-exchange capability-token minter and token store
(6/10 of EPIC #138), built on the unified `build_kv_store` factory from
#122 — no new storage abstraction.

- **`CapabilityTokenStore`** (`_file_exchange/_tokens.py`) with the
  `mint` / `lookup` / `consume` / `revoke` lifecycle. Expiry is delegated
  to the KV layer's TTL; single-use is enforced by an atomic `delete`
  whose boolean return gives at-most-once (§10.3). The per-token
  `metadata` is opaque — the store never interprets it; the #145/#146
  data planes own its shape, routes, and full-URL assembly.
- **`MintedToken` / `TokenRecord`** frozen dataclasses; `MintedToken.expires_at`
  reflects the *clamped* TTL so a route can build a descriptor's `expiresAt`
  directly. Tokens are 256-bit `secrets.token_urlsafe(32)` (§12 floor is 128-bit).
- **`build_capability_token_store(config)`** factory mirroring `build_event_store`,
  resolving the backend under `namespace="file-exchange-tokens"` and threading the
  TTL ceiling from `config.file_exchange_token_ttl`.
- **`capability_url(base_url, path, token)`** helper enforcing §12's `https`
  requirement (raises `ConfigurationError` for unset/non-https `base_url`).
- **Two `ServerConfig` env fields**: `file_exchange_token_ttl` (the ceiling
  `mint` clamps to) and `file_exchange_max_artifact_size` (surfaced here per
  #144's scope; consumed by the #146 upload route).
- **Public re-exports** of the five names across `_file_exchange/__init__.py`
  and the `file_exchange` namespace module.

Operator config (TTL ceiling, max size) is on the env axis, not kwargs;
the only `mint` kwargs (`ttl`, `single_use`) are per-call domain inputs.
Shape decisions (token entropy, namespace, https enforcement) are owned by
pvl-core. The design/plan docs live under `docs/superpowers/` (design
records, not normative wire specs).

## Test plan

- [ ] `uv run pytest` green on Python 3.10 and 3.13 (826 passed, 1 skipped — the GITHUB_TOKEN-gated spec-sync test)
- [ ] `uv run ruff format --check . && uv run ruff check . && uv run mypy src` clean
- [ ] Token lifecycle covered: entropy/URL-safety, TTL clamp (both directions),
      expiry boundary (lookup + consume), single-use invalidation, double-consume,
      multi-use no-op, revoke idempotency (incl. single-use), ceiling validation,
      and end-to-end via the config-driven factory
- [ ] `capability_url` covered: join/normalize/empty-path/empty-token/requires-base-url/requires-https
- [ ] Namespace re-export of all five names asserted in both `__all__`s

Closes #144


pvliesdonk:
Thanks @gemini-code-assist. After review I'm keeping the direct `record["metadata"]` / `record["single_use"]` subscripts in `lookup` and `consume` rather than switching to `.get()` with defaults. Reasoning:

1. **The malformed-record path is unreachable in normal operation.** `mint` is the sole writer of these records, and `build_capability_token_store` resolves the backend via `build_kv_store(config, namespace="file-exchange-tokens")`, which wraps it in a `PrefixCollectionsWrapper` and pins a dedicated `tokens` collection. A record lacking `metadata`/`single_use` can therefore only arise from out-of-band corruption or a foreign writer in the isolated keyspace.

2. **For a genuinely corrupt record, a loud `KeyError` is the correct failure — `.get()` with defaults would mask it.** The suggested `record.get("metadata", {}), record.get("single_use", True)` would fabricate a `TokenRecord` that looks like a *valid single-use token with empty metadata*, turning data corruption into a silently-usable capability. Likewise `consume`'s `record.get("single_use")` would be falsy on a corrupt record and return `True` (a successful no-op), falsely reporting consumption. For a capability-token primitive, surfacing corruption loudly is safer than coercing it into a plausible-looking valid token.

3. If we later want explicit defense-in-depth here, the correct shape is to treat a malformed record as **absent** (`lookup` → `None`, `consume` → `False`), not to fabricate defaults. I'll file that as a follow-up if it becomes a real concern rather than mask corruption now.

The earlier expires_at finding (compute before the `put`) was valid and is fixed in `f0b7e8d`.

PR #160, opened 2026-05-24
feat(file-exchange): SSRF + DNS-rebind guard for outbound HTTP

### LINKED ISSUE #147: file-exchange (9/10): SSRF + DNS-rebind guard for fetcher and sender outbound HTTP
Part of #138. Depends on the wire-format issue (1/10).

## Scope

Centralized outbound-HTTP helper used by the fetcher (download) and the sender (upload). Enforces the §15 SSRF mitigations and the §15 DNS-rebinding mitigation. Carries the URL-redaction discipline.

## Deliverables

- `https`-only enforcement.
- No-cross-origin-redirect behaviour.
- Refusal of loopback (`127.0.0.0/8`, `::1`), link-local (`169.254.0.0/16`, `fe80::/10`), and RFC 1918 ranges unless explicitly allowlisted via env config.
- Resolve-once-pin-IP: resolve the hostname once, connect to that pinned IP, prevents DNS rebinding between check and connection.
- URL redaction in every error path and every log line: capability URLs MUST NOT appear in logs (carries forward the discipline that gemini caught on PR #122).
- Unit tests covering each refusal case, the allowlist override, the pinned-IP behaviour, and the redaction.

## Spec references

§15 (security considerations — SSRF, DNS rebinding).

pvliesdonk:
## Summary

Adds the package-internal outbound-HTTP guard (`guarded_stream`) that the `download` fetcher (#145) and `upload` sender (#146) will issue every request through, enforcing the §15 SSRF + DNS-rebind mitigations. It is a thin §15 primitive — verification, `Range` recovery, and artifact handling stay with #145/#146.

- **Deny-all-non-global** address refusal (`ipaddress.is_global`, IPv4-mapped unwrapped) with an operator **CIDR allowlist** escape hatch.
- **Resolve-once-pin-IP** via httpx's `sni_hostname` extension: one `getaddrinfo`, connect to the validated IP, verify the cert against the hostname — no re-resolution between check and connect.
- **Redirects:** refuse cross-origin (and scheme-downgrade); follow same-origin bounded to 5 hops, **re-guarding each hop** (rebind-safe); refuse a redirect on a bodied request; non-redirect 3xx (`304`/no-`Location`) yielded as terminal.
- **No ambient credentials:** `trust_env=False`, fresh client per call, URL userinfo stripped, caller `Host` header stripped in favour of the validated hostname (non-default port preserved, IPv6 bracketed).
- **Redaction:** only the hostname reaches a log line; wire `detail` carries no URL parts; `GuardedResponse` keeps the response out of its `repr`.
- Two operator `ServerConfig` fields: `file_exchange_allowed_networks` (CIDRs) and `file_exchange_http_timeout` (default 30s). The guard stays **package-internal** — not re-exported through the public `file_exchange` surface.

Closes #147. Part of EPIC #138.

## Test plan

- [x] `uv run pytest` green on Python **3.10** and **3.13** (889 passed locally)
- [x] `uv run ruff format --check . && uv run ruff check . && uv run mypy src` clean
- [x] 57 guard tests cover: the deny-all-non-global matrix (loopback / link-local / RFC1918 / CGNAT / `0.0.0.0` / IPv4-mapped, verified on 3.10 + 3.13), CIDR allowlist, resolve-once-pin (Host + `sni_hostname`, IPv6 bracketing, non-default port), redirect policy (cross-origin / scheme-downgrade / rebind-to-blocked-IP / too-many / non-redirect-3xx / bodied-request), no-ambient-credentials (`trust_env`, caller-`Host` injection, URL-userinfo→`Authorization`), credential/token redaction across all four log emit-paths, and client cleanup on success / caller-exception / refusal / mid-body-error

### DIFF
--- a/docs/superpowers/plans/2026-05-24-file-exchange-147-ssrf-guard.md
+++ b/docs/superpowers/plans/2026-05-24-file-exchange-147-ssrf-guard.md
@@ -1,0 +1,910 @@
+# File-Exchange #147 — SSRF + DNS-rebind guard Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build the package-internal outbound-HTTP guard (`guarded_stream`) that the download fetcher (#145) and upload sender (#146) issue every request through, enforcing the §15 SSRF + DNS-rebind mitigations.
+
+**Architecture:** A thin async-context-manager primitive in a new module `_file_exchange/_outbound.py`. It enforces https-only, deny-all-non-global address refusal (with a CIDR allowlist), resolve-once-pin-IP via httpx's `sni_hostname` extension, cross-origin-redirect refusal (same-origin followed, bounded), and no-ambient-credentials. Every refusal/failure raises `FileExchangeTransferError(NOT_ACCESSIBLE, transport=<caller label>)`. Verification and `Range` recovery are NOT here — they belong to #145/#146.
+
+**Tech Stack:** Python 3.10+, `httpx` (already a dep), stdlib `ipaddress`/`socket`/`asyncio`, `pytest` + `pytest-asyncio` (`asyncio_mode = "auto"`), `httpx.MockTransport` for tests.
+
+**Design reference:** `docs/superpowers/specs/2026-05-24-file-exchange-147-ssrf-guard-design.md`.
+
+**No public surface change:** `guarded_stream`/`GuardedResponse` are imported by #145/#146 via `fastmcp_pvl_core._file_exchange._outbound`. They are NOT added to `file_exchange.py` or the subpackage `__init__.py` `__all__`. The only operator-facing additions are two `ServerConfig` fields.
+
+---
+
+### Task 1: ServerConfig fields
+
+**Files:**
+- Modify: `src/fastmcp_pvl_core/_config.py`
+- Test: `tests/test_config.py`
+
+- [ ] **Step 1: Write the failing tests**
+
+Add to `tests/test_config.py`:
+
+```python
+def test_file_exchange_outbound_defaults():
+    from fastmcp_pvl_core import ServerConfig
+
+    cfg = ServerConfig.from_env("MYAPP")
+    assert cfg.file_exchange_allowed_networks == ()
+    assert cfg.file_exchange_http_timeout == 30.0
+
+
+def test_from_env_file_exchange_allowed_networks(monkeypatch):
+    from fastmcp_pvl_core import ServerConfig
+
+    monkeypatch.setenv(
+        "MYAPP_FILE_EXCHANGE_ALLOWED_NETWORKS", "10.0.0.0/8, 192.168.5.0/24"
+    )
+    cfg = ServerConfig.from_env("MYAPP")
+    assert cfg.file_exchange_allowed_networks == ("10.0.0.0/8", "192.168.5.0/24")
+
+
+def test_from_env_file_exchange_http_timeout(monkeypatch):
+    from fastmcp_pvl_core import ServerConfig
+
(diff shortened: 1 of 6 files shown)


pvliesdonk:
## Review disposition

Thanks both. Dispositions below.

### gemini — `ipv4_mapped` "introduced in Python 3.12" (HIGH)

**This is a false positive — not applying.** `ipaddress.IPv6Address.ipv4_mapped` has existed since **Python 3.4**, not 3.12. Verified on the project's supported interpreters:

```
py 3.10.20  ipaddress.ip_address('::ffff:127.0.0.1').ipv4_mapped -> 127.0.0.1
py 3.11.14  ipaddress.ip_address('::ffff:127.0.0.1').ipv4_mapped -> 127.0.0.1
py 3.12.11  ipaddress.ip_address('::ffff:127.0.0.1').ipv4_mapped -> 127.0.0.1
```

CI corroborates: `test (3.10)` and `test (3.11)` pass, and the parametrized deny-matrix test exercises `::ffff:127.0.0.1` through `_is_permitted` on every version. The suggested hand-rolled `packed[:12] == b"\x00"*10 + b"\xff\xff"` rewrite would replace correct, readable stdlib usage with a manual equivalent for a non-existent incompatibility, so I'm keeping `.ipv4_mapped`.

### claude-review

- **#1 (Medium — IPv4-mapped CIDR in allowlist silently never matches): fixed** in `380d628`. `_parse_allowed_networks` now raises `ConfigurationError` for an IPv4-mapped IPv6 CIDR (via `subnet_of(::ffff:0:0/96)`, so a broad/ULA range like `fc00::/7` isn't over-rejected), plus tests.
- **#2 (body-vs-cross-origin check ordering, Low):** keeping body-first. A bodied request can't safely follow *any* redirect (the body iterator is consumed), so "refused redirect on a request with a body" is the more precise reason for that case; ordering it first is intentional.
- **#3 (`_same_origin` 443 default, by-design):** unreachable for non-https since `_send_pinned` enforces https before any redirect reaches `_same_origin`; the docstring notes the 443 normalization.
- **#4 (per-request CIDR parse, negligible):** intentional and documented in the design doc; `ServerConfig` is the dumb carrier and the consuming layer parses (consistent with the rest of the package).
- **#5 (no direct `_same_origin` unit test, Low):** covered indirectly by the redirect suite (same-origin follow, cross-origin/scheme-downgrade refusal, rebind-to-blocked); leaving as-is to avoid further diff churn.
- **#6 (912-line plan doc, Note):** `docs/superpowers/plans/` is the repo's standard home for per-issue implementation plans (every file-exchange child issue has one); keeping it consistent with the established convention.

Module-docstring summary also rounded out to mention the body-on-redirect refusal (`e530c5d`).

pvliesdonk:
## Round 4 disposition

CI green on 3.10–3.13; claude-review reports no security blockers. All three findings are LOW/Negligible:

- **#1 space-separated CIDR (LOW):** `FILE_EXCHANGE_ALLOWED_NETWORKS` is documented comma-separated; a space-separated value yields a malformed entry that already fails **loud** with `ConfigurationError`. Acceptable as-is — no silent misconfig.
- **#3 `suppress(Exception)` → `suppress(BaseException)` (Negligible): not applying — the suggested fix would regress cancellation.** `suppress(BaseException)` would swallow `asyncio.CancelledError`/`GeneratorExit`/`KeyboardInterrupt` during the discard-close, which breaks cooperative cancellation and shutdown. The current `suppress(Exception)` deliberately lets those `BaseException`s propagate; the only effect is a second `await response.aclose()` in the `finally`, which is a harmless no-op (httpx `aclose` is idempotent). Keeping `suppress(Exception)`.
- **#4 `float(http_timeout_str)` bare `ValueError` (Negligible):** pre-existing repo-wide `from_env` cast pattern (shared with `token_ttl`, `max_artifact_size`, `port`); tracked as #161 for a uniform fix rather than a one-off here.

No code change this round. PR is converged: the per-hop SSRF/rebind defense, redirect-`Location` handling (empty/malformed/cross-origin/body/cap), no-ambient-credentials, redaction, and the "all pre-yield failures are coded" contract are all verified and tested across Python 3.10–3.13.

PR #162, opened 2026-05-24
feat(file-exchange): download data plane — route + provider/fetcher (#145)

### LINKED ISSUE #145: file-exchange (7/10): download data plane — route + provider/fetcher

pvliesdonk:
## Summary

Implements the `download` transport's data plane (EPIC #138, 7/10), all streaming, never buffering an artifact whole in memory:

- **`download_provider_mint`** — mints a #144 capability token (`{"key": …}`) and returns a `TransferHandle` with a `DownloadSource` whose URL points at this server's own route. Lazy: the source hook is untouched at mint.
- **`register_file_exchange_routes`** — a FastMCP `GET <DOWNLOAD_PREFIX>/{token}` route that looks the token up (uniform `404`), serves the artifact via the #142 `ArtifactSource` hook on demand (streamed, `Range`/`206`/`416` supported), and consumes a single-use token only on a fully-retrieved unbounded request. Ambient credentials are ignored.
- **`download_fetcher_consume`** — retrieves through the #147 SSRF guard (`guarded_stream`) into a transient temp file (hashing while writing, `Range`-resuming on a drop), verifies size+digest **before** handing the sink a real sync fd (verify-before-use), then deletes the temp.

`DOWNLOAD_PREFIX` is a pvl-core shape constant (not a kwarg, not exported). The three role/route helpers are re-exported transport-qualified beside the `filesystem_*` ops.

## Key design decisions

- **Lazy serving** (route calls the hook per GET; no pvl-core-held copy). Eager staging was rejected — it would re-create a TTL'd blob store and contradict "stream end-to-end". Consequence: an `ArtifactSource` offered via `download` must yield stable bytes for the token lifetime; `digest` is whatever the offering metadata carries.
- **Temp-file fetcher** bridges the async `guarded_stream` to the synchronous `ArtifactSink` (which may read on the event loop) without deadlocking, and gives the sink the same real sync fd the `filesystem` transport does.
- **Status discipline**: initial GET must be `200`, a resume must be `206` (a Range-ignoring `200` would corrupt the running hash); a non-2xx is mapped to a §13 code, never ingested as artifact bytes.
- **Unknown size**: a `206` needs a `Content-Range` (RFC 9110 §15.3.7.1) which needs a total length, so when the hook does not report size the route ignores `Range` and serves `200`.

Design + plan: `docs/superpowers/specs/2026-05-24-file-exchange-145-download-data-plane-design.md`, `docs/superpowers/plans/2026-05-24-file-exchange-145-download-data-plane.md`.

## Test plan

- [x] Unit: provider mint shape; fetcher happy/digest-mismatch/size-mismatch (over+under)/too-large/guard-refusal/Range-resume/non-206-reject/reconnect-cap/temp-cleanup/local-write-error-not-retried/digest-only/empty-artifact/resume-corruption-caught.
- [x] Route: 404/200/206/416, Content-Length/Content-Range, single-use consume vs no-consume per range shape, multi-use repeat, ambient-credential ignore, unknown-size, open_artifact-failure → body-free 500, consume-failure logging, short-source no-consume.
- [x] Two-server e2e pull over a real ASGI app.
- [x] `uv run pytest` (945 passed) on Python 3.10 and 3.13; `ruff format --check`, `ruff check`, `mypy src` all clean.

Closes #145.


pvliesdonk:
Thanks for the reviews. Disposition of the findings:

**Fixed**
- **`tmp.flush()` OSError escaping (claude-review):** wrapped — a flush failure now maps to `transfer-failed` like the per-chunk write error, instead of a raw `OSError` breaking the `FileExchangeTransferError` contract. Added `test_fetcher_flush_error_maps_transfer_failed`.
- **Stream leak on a pre-handoff failure (gemini):** the route now closes the source stream if anything fails between `open_artifact` and the `StreamingResponse` taking ownership (an `except BaseException: close; raise` guard), mirroring the `os.fdopen` cleanup idiom.

**Defended (with changes to document the intent)**
- **TOCTOU on single-use tokens (claude-review):** this is intentional, spec-mandated behavior. §10.2 requires that *opening alone MUST NOT invalidate* (so `Range` resume can reconnect with the same token across drops). Consuming eagerly at lookup/open — or an atomic `take()` — would break resume and violate that clause. "single-use" here means *invalidated after the first full retrieval*, not a concurrency lock; with a bearer capability, any holder can fetch until consumption. I've documented this in the `register_file_exchange_routes` docstring and added `test_route_concurrent_single_use_both_served` to pin the behavior.
- **`Content-Disposition: attachment` (gemini):** this repo adopts `mcp-file-exchange-ext` **v0.1** (pinned commit `5f50a4e`), whose §10.2 download obligations are `Content-Type`/`Content-Length` (when known), Range, single-use, no ambient credentials — it does not mandate `Content-Disposition`. The "v0.5+" guidance referenced is a different/older spec lineage, not the one this server family implements, so I've not added the header.

Also filled local coverage gaps surfaced in review: closed-range `end`-clamp, and size-ok/digest-wrong ordering.

PR #163, opened 2026-05-26
feat(file-exchange): upload transport data plane (#146)

### LINKED ISSUE #146: file-exchange (8/10): upload data plane — route + receiver/sender

pvliesdonk:
## Summary
Adds the `upload` transport data plane (EPIC #138, 8/10): receiver mint, the PUT/POST serving route with verify-before-use, and the SSRF-guarded sender. Plus a behavior-preserving refactor that extracts `_staging.py` (shared digest/chunk primitives) and moves `register_file_exchange_routes` into `_routes.py` (the new cross-transport entry point), making `source` optional and adding `sink`/`config` for the upload branch.

**Three role primitives + one route:**
- `upload_receiver_mint(artifact_id, *, token_store, base_url, ttl, expected=None, method="PUT") -> IntakeTicket` — mint only, no hook runs.
- `register_upload_route(mcp, *, token_store, sink, config)` — PUT/POST `<UPLOAD_PREFIX>/{token}` route: streams the request body to a transient temp file (hashing), enforces `acceptMimeTypes` (RFC 7231 media-range), verifies `Content-Digest` (RFC 9530 + RFC 8941 member-param stripping), and `requireDigest` algorithm-list match — all **before** depositing through `sink.store_artifact`. Single-success-per-URL (consume only on success). Ambient credentials ignored. Body-free 4xx/5xx; temp deleted on every path.
- `upload_sender_consume(sink, source, key, *, config)` — stages to temp (single-pass SHA-256), pushes via `guarded_stream` with `Content-Length` / `Content-Digest` / optional `Content-Type`; non-2xx → `TRANSFER_FAILED`; guard refusal propagates coded.

**§148 boundary:** this PR ships the primitives + routes. Threading real `token_store`/`source`/`sink`/`config` from server bootstrap (and Tasks integration for large transfers) is #148.

**Concurrency contract documented:** the route docstring + the strengthened `ArtifactSink.store_artifact` contract together pin that two concurrent PUTs may both reach `store_artifact` (single-use is not a concurrency lock), and that sinks MUST finish reading the stream before returning.

## Test plan
- [x] `uv run pytest` — 993 passed, 1 skipped (GITHUB_TOKEN-gated spec sync); upload suite alone 44 + e2e
- [x] `uv run ruff format --check . && uv run ruff check . && uv run mypy src` — clean
- [x] Cross-version: `uv run --python 3.10/3.13 pytest tests/_file_exchange/test_upload*.py` — green on both
- [x] Five rounds of local `preflight-circus`; converged clean at the ≥80 bar
- [x] Coverage includes: receiver mint, helper unit (RFC 9530 / RFC 7231 / RFC 8941 edge cases incl. supported-malformed-no-fall-through, member-param stripping, `*/*` client-wildcard guard, multi-algo digest round-trip), route happy + replay-404 / 413 (operator-cap and ticket-cap, both directions) / 415 / 400 (digest mismatch + missing-required-digest + algo-list mismatch) / 500 (sink failure body-free) / consume-after-store-failure-still-204, temp-file unlinked on every failure path, flush+close OSError-mapping regression guards (mirrors merged #145 download tests), ambient-creds ignored, sender headers / non-2xx / guard refusal / temp-unlinked / close-after-flush-keeps-transfer-error, two-server push e2e.

Closes #146.

### DIFF
--- a/docs/superpowers/plans/2026-05-24-file-exchange-146-upload-data-plane.md
+++ b/docs/superpowers/plans/2026-05-24-file-exchange-146-upload-data-plane.md
@@ -1,0 +1,1606 @@
+# File-Exchange #146 — Upload Data Plane Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Implement the `upload` transport's push data plane — a receiver that mints a capability URL + the `PUT`/`POST` route that deposits the pushed bytes through the `ArtifactSink`, and a sender that pushes an artifact through the SSRF guard with an RFC 9530 `Content-Digest`.
+
+**Architecture:** Mirror of the merged #145 download plane. First refactor merged #145 code: extract shared digest/chunk primitives into `_staging.py`, and move the cross-transport route registrar `register_file_exchange_routes` into a new `_routes.py` (extracting `register_download_route` from `_download.py`). Then add `_upload.py` with `upload_receiver_mint`, the upload route, `upload_sender_consume`, and the RFC 9530 (`Content-Digest`) / RFC 7231 (media-range) helpers. The route streams the request body to a transient temp file (hashing + size-bounded), verifies digest/constraints **before** the sink sees the bytes, then consumes the single-use token only on a successful store.
+
+**Tech Stack:** Python 3.10–3.13, FastMCP (`@mcp.custom_route`), Starlette `Request`/`Response`, httpx (via `guarded_stream`), Pydantic v2 wire models, `pytest` + `pytest-asyncio`, `httpx.ASGITransport` for route/e2e tests.
+
+**Design doc:** `docs/superpowers/specs/2026-05-24-file-exchange-146-upload-data-plane-design.md`
+
+---
+
+## File Structure
+
+- **Create `src/fastmcp_pvl_core/_file_exchange/_staging.py`** — shared primitives moved out of `_download.py`: `_CHUNK`, `_HASHLIB_BY_LABEL`, `_digest_verifier`, `_write_chunk`. Imported by both `_download.py` and `_upload.py` (independent leaf modules; no circular import).
+- **Modify `src/fastmcp_pvl_core/_file_exchange/_download.py`** — import the four primitives from `_staging`; remove their local definitions and the now-unused `import hashlib`. Rename `register_file_exchange_routes` → `register_download_route` (signature `register_download_route(mcp, *, token_store, source)`; body unchanged).
+- **Create `src/fastmcp_pvl_core/_file_exchange/_routes.py`** — `register_file_exchange_routes(mcp, *, token_store, source=None, sink=None, config=None)`; mounts the download route iff `source`, the upload route iff `sink` (which requires `config`).
+- **Create `src/fastmcp_pvl_core/_file_exchange/_upload.py`** — `UPLOAD_PREFIX`, `upload_receiver_mint`, `register_upload_route`, `upload_sender_consume`, and the helpers `_format_content_digest`, `_parse_content_digest`, `_media_type_accepted`.
+- **Modify `src/fastmcp_pvl_core/_file_exchange/__init__.py`** and **`src/fastmcp_pvl_core/file_exchange.py`** — re-export `register_file_exchange_routes` from `._routes` (was `._download`); add `upload_receiver_mint`, `upload_sender_consume`.
+- **Create `tests/_file_exchange/test_upload.py`** — unit tests (receiver mint, helpers, route, sender).
+- **Create `tests/_file_exchange/test_upload_e2e.py`** — two-server push end-to-end.
+- **Modify `tests/test_file_exchange_namespace.py`** — assert the upload names are re-exported.
+
+---
+
+### Task 1: Extract `_staging.py` (refactor merged #145)
+
+This is a pure move refactor — no behaviour change. The existing download test suite is the regression gate.
+
+**Files:**
+- Create: `src/fastmcp_pvl_core/_file_exchange/_staging.py`
+- Modify: `src/fastmcp_pvl_core/_file_exchange/_download.py`
+- Test (regression): `tests/_file_exchange/test_download.py`
+
+- [ ] **Step 1: Create `_staging.py` with the moved primitives**
+
+```python
+"""Shared temp-file staging + digest primitives for the HTTP transports.
+
+The download fetcher (#145) and the upload route/sender (#146) both buffer a
+byte stream to a transient temp file with hashing and a size bound, and verify a
+declared digest. These primitives factor out the common parts. Each transport
+keeps its own read loop (download resumes via ``Range``; upload reads
+``request.stream()`` or a hook stream) and applies the
+``OSError -> transfer-failed`` mapping / cleanup-suppression contract around
+them.
+"""
+
+from __future__ import annotations
+
+import hashlib
+from typing import IO
+
+# Streaming chunk size for temp-file staging (1 MiB).
+_CHUNK = 1024 * 1024
+
+# Declared-digest label -> hashlib name; an unsupported label fails verification
+# (cannot verify -> digest-mismatch), never silently skips (§15).
+_HASHLIB_BY_LABEL = {"sha-256": "sha256", "sha-384": "sha384", "sha-512": "sha512"}
+
+
+def _digest_verifier(
+    declared: str | None,
+) -> tuple[hashlib._Hash | None, str | None, bool]:
+    """Return (hasher | None, expected_hex | None, unverifiable).
+
+    ``unverifiable`` is True when a digest is declared with an unsupported label
+    — verification must then fail (cannot verify), never silently skip (§15).
+    """
+    if declared is None:
+        return None, None, False
+    label, _, expected_hex = declared.partition(":")
+    name = _HASHLIB_BY_LABEL.get(label.lower())
+    if name is None:
+        return None, expected_hex, True
+    return hashlib.new(name), expected_hex.lower(), False
+
+
+def _write_chunk(tmp: IO[bytes], hasher: hashlib._Hash | None, chunk: bytes) -> None:
+    """Write a body chunk to the temp file and fold it into the running hash.
+
+    Both ops run off the event loop in a single ``asyncio.to_thread`` dispatch.
+    """
+    tmp.write(chunk)
+    if hasher is not None:
+        hasher.update(chunk)
+```
+
+- [ ] **Step 2: Point `_download.py` at `_staging`**
+
+In `src/fastmcp_pvl_core/_file_exchange/_download.py`:
+
+1. Delete the local definitions of `_CHUNK` (line ~54), `_HASHLIB_BY_LABEL` (line ~57), `_digest_verifier` (lines ~96–110), and `_write_chunk` (lines ~113–120). Leave `DOWNLOAD_PREFIX` and `_MAX_RECONNECTS` in place.
+2. Remove `import hashlib` (now unused in `_download.py`).
+3. Add this import alongside the other `_file_exchange` imports:
+
+```python
+from fastmcp_pvl_core._file_exchange._staging import (
+    _CHUNK,
+    _HASHLIB_BY_LABEL,
+    _digest_verifier,
+    _write_chunk,
+)
+```
+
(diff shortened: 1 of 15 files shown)


pvliesdonk:
**Re: claude-review round 3 Finding 2 (413 mid-stream may TCP-deadlock under load)** — keeping the current behavior intentionally; not draining the remaining body.

Rationale:
1. **Starlette/uvicorn handles the connection teardown:** the response is sent, the underlying ASGI server closes (or RSTs) the connection per HTTP semantics. RFC 7231 §6.5.11 explicitly endorses closing after a 413.
2. **Draining a hostile/oversize body is itself an attack surface:** a sender that sends a 10 GB body with a 1 KB cap would force the receiver to spend significant work *after* rejection. The early-return is the spec-aligned protection (`§15`: bounded reads on untrusted bytes).
3. **The trade-off is symmetric to the merged download data plane** which similarly aborts mid-stream on truncation/cap rather than draining.
4. **The bot scored this "Plausible" not "Confirmed"** and noted the framework-level mitigation.

If a senders-can't-see-413 issue surfaces in practice we can revisit with a bounded-drain (e.g. drain up to size_cap+chunk before returning), but absent evidence I'm preferring the early-abort.

pvliesdonk:
## Closing this PR — abandoning per CLAUDE.md after 4 bot rounds + 7 local circus rounds

**Decision rationale (CLAUDE.md, \`feedback_quality_over_fast_merge\` and the abandon-rule):** > "If a change has taken more than two or three rounds of review-driven fixes — local circus rounds or bot rounds — stop patching it. ... The right move is to discard the branch and re-implement the change in a single clean pass from \`main\`. The first attempt was a spike — it taught you the full shape of the change: every edge the reviewers surfaced, every constraint, every interaction."

This PR has run through 4 bot-review rounds and 7 local preflight-circus rounds. CI was green every round, all CODE lenses were clean every round, and each bot pass surfaced 3-4 new "Confirmed" findings on new angles. The bot's high-recall pass is genuinely inexhaustible at this rate; the diff has become a sediment of fixes-of-fixes (21 commits) and continued grinding gives diminishing returns. **Treating this as a spike that taught the full shape, abandoning, and reimplementing clean from main.**

The branch \`feat/146-upload-data-plane\` is retained for reference. A new clean PR will follow that bakes in all lessons below from day 1.

---

## Lessons learned — comprehensive baseline for the re-implementation

The re-implementation MUST encode all of the following from the first commit. They are cumulative findings across local-circus and bot rounds — applying them upfront produces one coherent diff instead of 21.

### Async/IO correctness
- **All temp-file \`close()\` calls in async functions** use \`await asyncio.to_thread(handle.close)\` (not sync \`handle.close()\`). Applies to \`tmp.close\`, \`ingest.close\`, and the \`_content()\` generator's read handle. Same convention as merged \`_download.py\`.
- **\`_content()\` annotation** is \`AsyncGenerator[bytes, None]\`, not \`AsyncIterator[bytes]\`. Matches the download fetcher.
- **\`os.fdopen\` failure paths** have BOTH \`except OSError\` (wrap to \`FileExchangeTransferError(TRANSFER_FAILED)\` for client-side, or \`Response(500)\` for server-side route) AND \`except BaseException\` (close fd + re-raise) — the second arm prevents fd leaks on \`CancelledError\`/\`KeyboardInterrupt\` between \`mkstemp\` and \`fdopen\`. Applies to BOTH the route handler and the sender; the route version was a recurring miss.
- **Sender's staging loop** must use FOUR distinct error scopes:
  1. \`try _write_chunk except OSError\` → \`TRANSFER_FAILED, detail="failed to stage..."\` (write-side IO)
  2. \`try tmp.flush except OSError\` → \`TRANSFER_FAILED, detail="failed to stage..."\` (flush-side IO)
  3. \`try ... except Exception\` (outer, after the above) → \`TRANSFER_FAILED, detail="artifact source read failed"\` (hook may raise anything; \`stream.read\` OSError must NOT be conflated with \`_write_chunk\` OSError)
  4. \`try guarded_stream(...) except OSError\` → \`TRANSFER_FAILED, detail="failed to read the staged artifact during upload"\` (\`_content()\` open/read failures surfacing through httpx)

### Route handler (\`_serve_upload\`) error structure
- **\`request.stream()\` iteration is OUTSIDE the \`except OSError\`** — Starlette's \`ClientDisconnect\` is not an \`OSError\` and propagates cleanly to the framework when wrapped this way. Only the per-chunk \`_write_chunk\` and \`tmp.flush\` calls need OSError → \`Response(500)\` mapping.
- **\`store_artifact\` failure** returns body-free \`Response(500)\` with \`logger.exception\` — never echo the hook's exception message to the wire. Sink contract is \`raise on failure\`, no exception-type guarantee.
- **Token \`consume\` failure after a successful store** is logged and the route returns 204 anyway (data already landed; failing the request would mislead the client). The log must be \`logger.exception\` (not \`warning\`) to capture the kv-backend traceback; include \`artifact_id\` context.
- **Temp-file cleanup** uses outer \`try/finally: await asyncio.to_thread(os.unlink, tmp_path)\` with \`contextlib.suppress(OSError)\` — guarantees unlink on every exit path (404/413/415/400/500/204).
- **Mid-stream 413** returns immediately without draining the rest of the body — Starlette/uvicorn handles the connection teardown per RFC 7231 §6.5.11. Don't drain (it's an attack vector for hostile-oversize uploads — \`§15\` bounded-read principle).

### \`requireDigest\` semantics
- The route's \`requireDigest\` check enforces \`cd[0] in {a.lower() for a in expected.requireDigest}\` — algorithm-list membership, not just presence. \`_parse_content_digest\` returns the FIRST supported member; senders that include multiple members must place a required algorithm first.
- The sender accepts an optional \`digest_algo: str = _DEFAULT_DIGEST_LABEL\` parameter so callers can match arbitrary receiver \`requireDigest\` constraints. Validate it against \`_HASHLIB_BY_LABEL\` at function entry (\`ValueError\` if not in the supported set).

### \`Content-Digest\` parser (\`_parse_content_digest\`)
- Strip RFC 8941 member parameters before the colon-wrap check (\`value.split(";", 1)[0].strip()\`) — conformant senders may decorate digest members.
- A SUPPORTED-algorithm member with malformed bytes (bad colon-wrap or invalid base64) returns \`None\` immediately — never falls through to a later member (\`§15\`: never silently skip).
- The byte-sequence length guard must be \`len(value) <= 2\` (not \`< 2\`) — \`"::"\` is a structurally valid 2-char string that decodes to \`b""\` and would otherwise admit a zero-length "digest".

### Digest comparison
- Use \`hmac.compare_digest(hasher.digest(), cd[1])\` (constant-time), AND pre-check \`len(cd[1]) == hasher.digest_size\` — Python docs explicitly note \`compare_digest\` with mismatched-length inputs is not constant-time.
- The route's recorded \`ArtifactMetadata.digest\` uses \`label:hex\` form (\`f"{algo_label}:{hexdigest}"\`); the wire \`Content-Digest\` header uses RFC 9530 byte-sequence form (\`label=:base64:\`). These are distinct formats; \`_parse_content_digest\` and \`_format_content_digest\` handle the header form, \`_digest_verifier\` handles the metadata form.

### Module structure (\`_staging.py\` extraction)
- \`_HASHLIB_BY_LABEL\` is shared by ALL transports (download verify, filesystem verify, upload Content-Digest algorithm validation).
- \`_CHUNK\` is the canonical chunk size for buffered reads across ALL transports (download fetcher, upload route/sender, filesystem reader).
- \`_digest_verifier\` is download-only (filesystem inlines equivalent logic; upload uses RFC 9530 path).
- \`_write_chunk\` is HTTP-transport-only (download fetcher + upload route/sender for buffered staging).

### \`register_file_exchange_routes\` (\`_routes.py\`)
- Validate ALL preconditions BEFORE any \`register_*_route\` call — atomicity invariant. The sequence must be:
  ```python
  if source is None and sink is None: raise ValueError("...source/sink required...")
  if sink is not None and config is None: raise ValueError("...config required for upload...")
  if source is not None: register_download_route(...)
  if sink is not None: register_upload_route(...)
  ```
  A partial mount (download succeeds, ValueError raised before upload) violates the all-or-nothing contract.

### Hook contract (\`_hooks.py\` \`ArtifactSink.store_artifact\`)
- The protocol docstring MUST state: "The caller closes \`stream\` and may delete its backing storage immediately after this method returns, so the sink MUST finish reading before returning (it may read on the loop or off-load to a thread, but MUST NOT retain the handle for deferred reads)."
- The route's docstring MUST state the concurrent-PUT contract: "single-use" means burn-on-success, not a concurrency lock; two concurrent PUTs can both reach \`store_artifact\` before either \`consume\` wins; the \`ArtifactSink\` MUST tolerate duplicate \`store_artifact\` calls for the same \`artifact_id\` within the TTL window.

### Test coverage requirements (regression guards)
- Temp-file leak on every failure path (route 404/413/415/400/500 + sender non-2xx/guard-refusal) — \`monkeypatch.setattr(tempfile, "mkstemp", capturing_mkstemp)\` + \`assert not os.path.exists(p)\`.
- Flush failure → \`Response(500)\`; close failure after success → still 204 (\`contextlib.suppress(OSError)\` preserves the success path); close after flush failure keeps the in-flight \`FileExchangeTransferError\` (sender).
- \`tmp.close\` and \`ingest.close\` close-via-to_thread — patterned via a wrapping fdopen.
- Consume-after-store failure → route still returns 204.
- requireDigest algorithm enforcement: missing-header-when-required → 400; algorithm-mismatch → 400; multi-algo accept; value-mismatch → 400.
- size-cap minimum-of-both (operator cap and ticket cap, both directions).
- sha-256/384/512 end-to-end (route hashes with the inbound algo, records metadata with the same algo).
- Ambient credentials ignored (Authorization/Cookie headers don't bypass).
- Two-server push e2e via \`httpx.ASGITransport\` (mock \`guarded_stream\` to route to the receiver's app).
- Sender hook-stream non-OSError wrapped as \`TRANSFER_FAILED\`.
- \`register_file_exchange_routes\` requires source-or-sink + requires config-when-sink (separate tests for each branch).

### Lens-attribution lessons (for re-implementation's local circus)
- **Cross-file API contract mismatches** (sender capability vs receiver requirement) have no local lens that catches them — sender's \`digest_algo\` gap was found by bot, not circus.
- **Operator-ergonomics / debuggability** items (artifact_id-in-logs, no-cancellation-log) consistently Haiku-score sub-80 but bots flag them. Either bake them in upfront or accept the residual.
- **Pre-existing pvl-core conventions** (broad-Exception suppress on hook streams, sync close on temp files) are NOT pre-existing — they were download-fetcher mistakes that #145 happens to share. The local circus's "matches the merged \`_download.py\` convention" defense was wrong on multiple items; treat \`_download.py\` as a reference but verify each pattern against the actual cleanest shape.
- **Test coverage symmetry with download tests** is consistently ≥80 — when adding any \`_upload.py\` error path, write the analogous test that the download fetcher has.

### Items the spike resolved that should NOT be re-debated
- \`Content-Disposition\` header on the download response: not required by spec; the previous PR #162 defended this.
- 4xx-vs-5xx granular mapping in the sender: spec-aligned to collapse to \`TRANSFER_FAILED\`; the design doc chose this shape.
- Sync vs async \`stream.close\` on hook streams: \`contextlib.suppress(Exception)\` is the documented pvl-core stance (hooks are untrusted downstream code).

---

## What remains valid

- **Design doc** (\`docs/superpowers/specs/2026-05-24-file-exchange-146-upload-data-plane-design.md\`) — the design is sound; lessons above are precision additions on top of it.
- **Plan** (\`docs/superpowers/plans/2026-05-24-file-exchange-146-upload-data-plane.md\`) — task structure is fine; individual task implementations should incorporate the lessons.
- **Tests** in this branch — most are valuable regression guards; copy them across in the re-implementation.
- **Branch** \`feat/146-upload-data-plane\` retained as a reference spike (do not delete).

## Next step

A new clean PR for #146 will be opened from a fresh branch off \`main\`, in a separate session, applying all lessons above from the first commit. Issue #146 stays open.

PR #164, opened 2026-05-26
feat(file-exchange): upload transport data plane (#146)

pvliesdonk:
## Summary
Clean re-implementation of #146 (upload transport data plane), bringing the EPIC #138 cleanroom rewrite to 8/10. Adds receiver mint, the PUT/POST serving route with verify-before-use, and the SSRF-guarded sender. Plus a behaviour-preserving refactor that extracts `_staging.py` (shared digest/chunk primitives) and moves `register_file_exchange_routes` into `_routes.py` (the cross-transport entry point), and consolidates `_filesystem.py`'s `_HASHLIB_BY_LABEL`/`_CHUNK` into `_staging.py` to eliminate silent-drift risk.

**Three role primitives + the route:**
- `upload_receiver_mint(artifact_id, *, token_store, base_url, ttl, expected=None, method="PUT") -> IntakeTicket` — mint only, no hook runs.
- `register_upload_route(mcp, *, token_store, sink, config)` — PUT/POST `<UPLOAD_PREFIX>/{token}`: streams the request body to a transient temp file (hashing), enforces `acceptMimeTypes` (RFC 7231 media-range) and `requireDigest` (algorithm-list membership; first-supported semantics documented), verifies `Content-Digest` (RFC 9530 + RFC 8941 member-param stripping; rejects empty `\"::\"` payloads; `hmac.compare_digest` with length pre-check for constant-time mismatched-length handling) **before** depositing through `sink.store_artifact`. Single-success-per-URL (consume only on success). Ambient credentials ignored. Body-free 4xx/5xx; temp deleted on every path. `request.stream()` `ClientDisconnect` propagates to Starlette unwrapped (narrow `except OSError` covers only `_write_chunk` and `tmp.flush`).
- `upload_sender_consume(sink, source, key, *, config, digest_algo="sha-256")` — stages to temp (single-pass hash with caller-selectable algorithm to match arbitrary receiver `requireDigest`), pushes via `guarded_stream` with `Content-Length` / `Content-Digest` / optional `Content-Type`; non-2xx → `TRANSFER_FAILED`; guard refusal propagates coded; hook-stream non-OSError wrapped as `TRANSFER_FAILED` with `\"artifact source read failed\"` detail (distinct from `_write_chunk`/`tmp.flush` OSErrors which use `\"failed to stage…\"`).

[...]

PR #165, opened 2026-05-26
feat(file-exchange): upload transport data plane (#146)

### LINKED ISSUE #146: file-exchange (8/10): upload data plane — route + receiver/sender

pvliesdonk:
## Summary
Clean re-implementation of #146 (upload transport data plane), 8/10 of EPIC #138. Adds `upload_receiver_mint`, the PUT/POST serving route, and the SSRF-guarded sender. Plus a behaviour-preserving refactor extracting `_staging.py` (shared digest/chunk primitives, consolidated across all three transports including `_filesystem.py`) and moving `register_file_exchange_routes` into a new `_routes.py` (cross-transport entry point with strict precondition validation before any mount).

## Replaces #163 + #164
PR #163 was a spike that ran 4 bot-review rounds + 7 local circus rounds without converging (21 commits). PR #164 was a consolidated push of the spike's final state that continued the spiral. Both abandoned per CLAUDE.md. This PR is the **true** clean re-implementation: written fresh from the design + plan + cumulative findings, in 4 coherent commits, with every lesson baked in from commit 1.

The cumulative lessons (from spike + v2 bot rounds) are encoded as concrete patterns in the code:
- **Async correctness:** all `close()` calls via `await asyncio.to_thread(...)`; `os.fdopen` has BOTH `except OSError` and `except BaseException` arms in route AND sender (fd-leak guard under cancellation); route's `except OSError` narrowed to wrap only `_write_chunk`/`tmp.flush` so Starlette's `ClientDisconnect` propagates cleanly.
- **Sender error contract:** wraps `source.open_artifact`; catches `httpx.HTTPError` alongside `OSError` around `guarded_stream` (network errors mapped to `TRANSFER_FAILED`); staging loop uses separate inner scopes for `_write_chunk`/`tmp.flush` with the outer scope catching hook-side errors as `\"artifact source read failed\"`; `except FileExchangeTransferError: raise` re-raise prevents inner FETE from being re-wrapped.
- **Digest verification:** RFC 9530 `Content-Digest` parser strips RFC 8941 member parameters, rejects `\":\":\"` empty digests (`len(value) <= 2` guard), supported-but-malformed returns `None` (never silently skips per §15); `hmac.compare_digest` with `len(cd[1]) != hasher.digest_size` length pre-check for constant-time mismatched-length handling.
- **`requireDigest` enforcement:** set-membership check on the supplied algorithm; first-supported-member semantics documented in the route docstring.
- **Sender `digest_algo` parameter:** lets callers match arbitrary receiver `requireDigest` constraints (default `sha-256`); `ValueError` on unsupported algorithm.
- **`_content()` generator:** annotated `AsyncGenerator[bytes, None]`; explicit `aclose()` in a finally for Windows safety + cleaner POSIX semantics.
- **`register_file_exchange_routes` atomicity:** validates ALL preconditions BEFORE any `register_*_route` call; no partial-mount-then-raise.
- **`ArtifactSink.store_artifact` contract strengthened** with caller-closes-and-deletes-after-return wording.
- **Concurrency contract documented** in route docstring: single-use is not a lock; sinks MUST tolerate duplicate `store_artifact` calls within TTL.
- **`_filesystem.py`** consolidated to import `_HASHLIB_BY_LABEL` AND `_CHUNK` from `_staging.py` (no silent-drift opportunity).

## Test plan
- [x] `uv run pytest` — 1023 passed, 1 skipped (GITHUB_TOKEN-gated spec sync)
- [x] `uv run ruff format --check . && uv run ruff check . && uv run mypy src` — clean
- [x] Cross-version: `uv run --python 3.10/3.13 pytest tests/_file_exchange/test_upload*.py` — 64 passed on both
- [x] Coverage includes (regression guards from spike+v2 lessons): receiver mint, helper unit (RFC 9530 / 7231 / 8941 edge cases incl. `\":\":\"` rejection, member-param stripping, supported-malformed-no-fall-through, multi-algo digest round-trip), route status matrix (204/404/413/415/400/500), `requireDigest` algo-list enforcement, `digest_algo` happy + ValueError on unsupported algo, ambient-creds ignored, consume-failure-still-204, temp-file unlinked on every failure path (route + sender), flush+close error-mapping mirroring merged #145 download tests, sender hook-stream non-OSError → TRANSFER_FAILED, `register_file_exchange_routes` source-or-sink + config-when-sink + atomicity (download NOT mounted when config-missing ValueError fires), two-server push e2e.

Closes #146.

### DIFF
--- a/docs/superpowers/plans/2026-05-24-file-exchange-146-upload-data-plane.md
+++ b/docs/superpowers/plans/2026-05-24-file-exchange-146-upload-data-plane.md
@@ -1,0 +1,1606 @@
+# File-Exchange #146 — Upload Data Plane Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Implement the `upload` transport's push data plane — a receiver that mints a capability URL + the `PUT`/`POST` route that deposits the pushed bytes through the `ArtifactSink`, and a sender that pushes an artifact through the SSRF guard with an RFC 9530 `Content-Digest`.
+
+**Architecture:** Mirror of the merged #145 download plane. First refactor merged #145 code: extract shared digest/chunk primitives into `_staging.py`, and move the cross-transport route registrar `register_file_exchange_routes` into a new `_routes.py` (extracting `register_download_route` from `_download.py`). Then add `_upload.py` with `upload_receiver_mint`, the upload route, `upload_sender_consume`, and the RFC 9530 (`Content-Digest`) / RFC 7231 (media-range) helpers. The route streams the request body to a transient temp file (hashing + size-bounded), verifies digest/constraints **before** the sink sees the bytes, then consumes the single-use token only on a successful store.
+
+**Tech Stack:** Python 3.10–3.13, FastMCP (`@mcp.custom_route`), Starlette `Request`/`Response`, httpx (via `guarded_stream`), Pydantic v2 wire models, `pytest` + `pytest-asyncio`, `httpx.ASGITransport` for route/e2e tests.
+
+**Design doc:** `docs/superpowers/specs/2026-05-24-file-exchange-146-upload-data-plane-design.md`
+
+---
+
+## File Structure
+
+- **Create `src/fastmcp_pvl_core/_file_exchange/_staging.py`** — shared primitives moved out of `_download.py`: `_CHUNK`, `_HASHLIB_BY_LABEL`, `_digest_verifier`, `_write_chunk`. Imported by both `_download.py` and `_upload.py` (independent leaf modules; no circular import).
+- **Modify `src/fastmcp_pvl_core/_file_exchange/_download.py`** — import the four primitives from `_staging`; remove their local definitions and the now-unused `import hashlib`. Rename `register_file_exchange_routes` → `register_download_route` (signature `register_download_route(mcp, *, token_store, source)`; body unchanged).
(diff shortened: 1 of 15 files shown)


pvliesdonk:
Closing — third bot-iteration round, the cycle is not converging. Per the CLAUDE.md abandon-rule (and the recently-saved memory \`feedback_high_recall_bot_iteration_limits\`), the right move at this point is not "one more fix round" but to recognize that **#146's scope is too large for a single PR**.

Evidence:
- PR #163 (spike): 4 bot rounds + 7 local circus rounds, never converged (21 commits).
- PR #164 (consolidated push): repeated the pattern — same code structure, same iteration spiral.
- PR #165 (true clean re-implementation, written fresh): 3 bot rounds, each producing real-but-narrow findings (fd-leak windows, atomicity wording, parser edges, error-type narrowing). CI green throughout. The findings keep coming because the diff surface is large.

The first two abandons read as "fix the implementation approach." The third abandon reads as "fix the scope." The design + plan + first-iteration tests all remain valid; only the "single PR" packaging needs to change.

The right next step is a brainstorming session to decompose #146 into smaller PRs (refactor / receiver-side / route / sender / e2e — each landing independently with a small review surface that converges in one round). That brainstorming is a fresh session's work; not bolted onto this drift.

Branch \`feat/146-upload-data-plane-v3\` retained as a third-attempt reference. Issue #146 stays open and needs decomposition before re-attempting.

PR #169, opened 2026-05-28
feat(file-exchange): upload PUT/POST route + edge modes (slice 3/5 of #146)

pvliesdonk:
## Summary

Third of 5 stacked PRs for #146. Builds on #168 (mint + helpers). Adds \`register_upload_route\` — the upload data plane's serving route — plus a comprehensive edge-mode test sweep covering matrix rows A1–A6, B1/B4/B6, C1/C2, D4–D8, F4–F5.

## Route behaviour

\`PUT\`/\`POST /fx/u/{token}\`:

- Token lookup miss → 404 (uniform for unknown / expired / consumed; no token state leakage)
- \`acceptMimeTypes\` mismatch → 415 (no consume, no sink call)
- Effective body cap = \`min(expected.maxSize, config.file_exchange_max_artifact_size)\` → over cap → 413 with \`Connection: close\` (prevents next-request misframing on HTTP/1.1 keep-alive)
- \`Content-Digest\` verification BEFORE the sink sees the bytes:
  - missing-when-\`requireDigest\` → 400
  - present-but-unparseable → 400
  - present-but-unsupported algorithm → 400
  - mismatch → 400
  - \`requireDigest\` enforces algorithm-list membership, not just presence
- Sink raise → 500 (token NOT consumed; subsequent PUT can retry)
- \`token_store.consume\` runs only AFTER successful sink return (matrix row A1+A2 contract)
- \`starlette.requests.ClientDisconnect\` caught and re-raised (it inherits from \`Exception\`, not \`OSError\`)
- Temp file unlinked on every path

## Test plan

- [x] \`uv run pytest tests/_file_exchange\` — 613 passed (= 588 from slice 2 + 25 new route tests)
- [x] \`uv run --python 3.10 pytest tests/_file_exchange\` — 613 passed
- [x] \`uv run --python 3.13 pytest tests/_file_exchange\` — 613 passed
- [x] \`uv run ruff format --check .\` clean

**Base:** depends on #168 → #167. Merge those first.

Part of #146.

### DIFF
--- a/src/fastmcp_pvl_core/_file_exchange/_staging.py
+++ b/src/fastmcp_pvl_core/_file_exchange/_staging.py
@@ -58,3 +58,21 @@ def _write_chunk(tmp: IO[bytes], hasher: hashlib._Hash | None, chunk: bytes) -> 
     tmp.write(chunk)
     if hasher is not None:
         hasher.update(chunk)
+
+
+def _rehash_file(path: str, hashlib_name: str) -> bytes:
+    """Synchronously compute ``hashlib_name``'s digest of the file at ``path``.
+
+    Designed to be called via a single ``asyncio.to_thread`` dispatch
+    from the upload route's non-sha-256 verify branch — the entire
+    open + chunked-read + close runs in one worker thread rather than
+    spawning a thread per chunk.
+    """
+    h = hashlib.new(hashlib_name)
+    with open(path, "rb") as fh:
+        while True:
+            buf = fh.read(_CHUNK)
+            if not buf:
+                break
+            h.update(buf)
+    return h.digest()
--- a/src/fastmcp_pvl_core/_file_exchange/_upload.py
+++ b/src/fastmcp_pvl_core/_file_exchange/_upload.py
@@ -36,8 +36,8 @@ from typing import TYPE_CHECKING, Literal, cast
 
 from fastmcp_pvl_core._file_exchange._spec import SPEC_VERSION, TICKET_TYPE
 from fastmcp_pvl_core._file_exchange._staging import (
-    _CHUNK,
     _HASHLIB_BY_LABEL,
+    _rehash_file,
     _write_chunk,
 )
 from fastmcp_pvl_core._file_exchange._tokens import capability_url
@@ -214,24 +214,6 @@ def _media_range_matches(content_type: str, accept: list[str]) -> bool:
         ):
             return True
     return False
-
-
-def _rehash_file(path: str, hashlib_name: str) -> bytes:
-    """Synchronously compute ``hashlib_name``'s digest of the file at ``path``.
-
-    Called via a single ``asyncio.to_thread`` dispatch from the route's
-    rehash branch (the non-sha-256 ``Content-Digest`` verify path), so the
-    entire open + chunked-read + close runs in one worker thread rather
-    than spawning a thread per chunk.
-    """
-    h = hashlib.new(hashlib_name)
-    with open(path, "rb") as fh:
-        while True:
-            buf = fh.read(_CHUNK)
-            if not buf:
-                break
-            h.update(buf)
-    return h.digest()
 
 
 def register_upload_route(
--- a/src/fastmcp_pvl_core/_file_exchange/_upload.py
+++ b/src/fastmcp_pvl_core/_file_exchange/_upload.py
@@ -214,6 +214,7 @@ def register_upload_route(
     body-size cap; per-mint ``expected.maxSize`` is the smaller of the two when
     set. See the failure-mode matrix for the full per-status-code contract.
     """
+    from starlette.requests import ClientDisconnect
     from starlette.responses import Response
 
     async def _handle(request: Request) -> Response:
@@ -273,6 +274,11 @@ def register_upload_route(
                             break
                         await asyncio.to_thread(_write_chunk, tmp, hasher, chunk)
                         received += len(chunk)
+                except ClientDisconnect:
+                    # Client dropped mid-upload — no response needed; outer
+                    # finally still unlinks the temp.
+                    logger.debug("file-exchange: upload client disconnected mid-stream")
+                    raise
                 except OSError:
                     logger.exception("file-exchange: upload temp write failed")
                     return Response(status_code=500)
@@ -281,7 +287,7 @@ def register_upload_route(
                     await asyncio.to_thread(tmp.close)
 
             if too_large:
-                return Response(status_code=413)
+                return Response(status_code=413, headers={"Connection": "close"})
 
             cd_header = request.headers.get("content-digest")
             required = expected.requireDigest if expected is not None else None
@@ -300,12 +306,16 @@ def register_upload_route(
                 else:
                     rehash = hashlib.new(_HASHLIB_BY_LABEL[cd_algo])
                     try:
-                        with open(tmp_path, "rb") as fh:
+                        fh = await asyncio.to_thread(open, tmp_path, "rb")
+                        try:
                             while True:
                                 buf = await asyncio.to_thread(fh.read, _CHUNK)
                                 if not buf:
                                     break
                                 rehash.update(buf)
+                        finally:
+                            with contextlib.suppress(OSError):
+                                await asyncio.to_thread(fh.close)
                     except OSError:
                         logger.exception("file-exchange: upload rehash read failed")
                         return Response(status_code=500)
--- a/src/fastmcp_pvl_core/_file_exchange/_upload.py
+++ b/src/fastmcp_pvl_core/_file_exchange/_upload.py
@@ -268,11 +268,22 @@ def register_upload_route(
             return Response(status_code=404)
         artifact_id = cast("str", rec.metadata["artifact_id"])
         expected_raw = rec.metadata.get("expected")
-        expected = (
-            ArtifactConstraints.model_validate(expected_raw)
-            if expected_raw is not None
-            else None
-        )
+        try:
+            expected = (
+                ArtifactConstraints.model_validate(expected_raw)
+                if expected_raw is not None
+                else None
+            )
+        except Exception:
+            # KV-stored metadata that fails Pydantic validation should
+            # never happen (the mint side validated before storing) but
+            # KV corruption / schema-version drift can produce it. Log
+            # and return a structured 500 rather than letting a
+            # ValidationError escape unwrapped to ASGI.
+            logger.exception(
+                "file-exchange: upload expected-constraints deserialise failed"
+            )
+            return Response(status_code=500)
 
         content_type = request.headers.get("content-type", "")
         if expected is not None and expected.acceptMimeTypes is not None:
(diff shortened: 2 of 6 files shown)


pvliesdonk:
## Response to bot reviews on PR #169

Five of seven findings applied in 1f8b1b2. One rejected, one applied as a different fix than the reviewer suggested. Full breakdown:

### Applied

| # | Severity | Fix |
|---|---|---|
| 1 | HIGH (CONFIRMED) | \`_content_digest_parse\` gained a \`preferred=\` kwarg; route threads \`preferred=required\` so the requireDigest algorithm is selected regardless of header entry order. Regression test \`test_route_require_digest_picks_required_algo_regardless_of_order\` verifies \`sha-512=:...:, sha-256=:...:\` with \`requireDigest=[sha-256]\` now returns 204. |
| 3 | TEST GAP | C1 concurrent test now asserts \`1 <= len(sink.calls) <= 2\` after the gather. |
| 4 | CONTRACT GAP | Consume-failure warning message rewritten to flag the double-deposit risk on retry; added an explicit code comment distinguishing at-most-once-delivery from at-most-once-deposit. |
| 5 | CLAUDE.md | \`register_upload_route\` docstring now labels every kwarg with its category — \`token_store\` (shape), \`sink\` (hook), \`config\` (config). |
| 6 | PERFORMANCE | Extracted \`_rehash_file(path, hashlib_name)\` synchronous helper; single \`asyncio.to_thread\` dispatch replaces per-chunk \`await\` calls. |

### Rejected: Finding 2 — \`meta.digest\` "wrong algorithm"

\`ArtifactMetadata.digest\` is recorded in pvl-core's \`sha-256:<hex>\` form regardless of which algorithm the client attested via \`Content-Digest\`. This is the spec design — verify-against-attested-algo and record-in-pvl-core-form are deliberately decoupled. The route always computes sha-256 incrementally during the body read; when the client attests a non-sha-256 algorithm, a separate rehash verifies *their* attestation, but the *record* stays sha-256 so downstream consumers reading \`meta.digest\` have a stable canonical form. The reviewer's concern ("downstream re-verifying via meta.digest will compare the wrong algorithm") doesn't apply — \`meta.digest\`'s algorithm prefix is the truth about what's stored, not what was attested.

### Treated as a leftover, not added: Finding 7 — POST not tested

Slice 2 (#168 merged) made the route method a pvl-core constant (\`_UPLOAD_METHOD: Literal["PUT"] = "PUT"\`) — POST is never minted by pvl-core anymore. The route declaring \`methods=["PUT", "POST"]\` was leftover dead surface. **Dropped POST** from the route's \`methods=[...]\` rather than adding a POST test for a method nothing will ever ask for. (If a future downstream genuinely needs POST, per CLAUDE.md that's a spec evolution, not an override.)

All checks green on 3.10 + 3.13; 620 tests in the file-exchange suite.

pvliesdonk:
## Response to third bot pass on 8a0eb0c

### Applied on this slice (commit 8a0eb0c)

| # | Severity | Fix |
|---|---|---|
| 2 | TEST GAP (CONFIRMED) | New \`test_route_non_sha256_digest_mismatch_400\` pins the rehash mismatch branch — sha-512 required, header is well-formed but for the wrong bytes -> 400, no sink call |
| 3 | CONFIRMED | \`ArtifactConstraints.model_validate\` now wrapped in \`try/except\`; KV-corrupt metadata yields a structured 500 with \`logger.exception\` rather than letting \`ValidationError\` escape to ASGI |
| 4 | CONFIRMED LOW | Captured \`consume\` return value; on \`False\` (race-loser or revoke-after-store) logs at INFO with the "bytes are in the sink" framing — observability only |

### Routed to a separate PR

Finding 1 (\`requireDigest=[""]\` validation gap) lives in \`_wire.py\`'s \`ArtifactConstraints\`, not in the route. Filed as **#174** so slice 3 stays narrow; the chain will inherit on rebase. Same packaging discipline as #173 was for the post-#167 \`_digest_verifier\` fixes.

### Rejecting Gemini's "accumulate to 64KB" suggestion

The upload route's chunk loop is **deliberately symmetric** with the merged download fetcher:

\`\`\`python
# _download.py (on main, slice 1.5 territory)
async for chunk in resp.aiter_bytes():
    ...
    await asyncio.to_thread(_write_chunk, tmp, hasher, chunk)
\`\`\`

\`\`\`python
# _upload.py (this PR)
async for chunk in request.stream():
    ...
    await asyncio.to_thread(_write_chunk, tmp, hasher, chunk)
\`\`\`

Per-chunk dispatch is the pattern the download data plane shipped with (PR #162, merged) and the choice was discussed during the spec design. Introducing a 64KB accumulation buffer on the upload side would:

1. **Create asymmetry** between the two HTTP transports for no measurable benefit at typical chunk sizes (httpx and h11 already deliver chunks in the 16-64KB range under default settings, so a 64KB accumulator's amortisation factor is ≈1).
2. **Add per-chunk memory pressure** (the \`bytearray\` grows up to 64KB above what the OS pipe buffer already holds) on the receiver side, where one of pvl-core's hard constraints is that the receiver streams to disk without retaining a whole-artifact buffer.
3. **Break the matrix's "every temp-file op maps to transfer-failed or is suppressed for cleanup" invariant** (the per-chunk \`to_thread\` is where the \`OSError -> transfer-failed\` mapping lives; folding writes into a larger flush adds a separate error site to maintain).

If a future profile shows the per-chunk overhead is actually load-bearing under small-chunk clients, that's a transport-wide change applied to both upload AND download via \`_staging.py\`, not a one-sided introduction on upload.

Branch is at 8a0eb0c; 622 file-exchange tests pass on 3.10 + 3.13.

pvliesdonk:
## Pausing #169 to address the root-cause pattern

Across rounds 1–3 of bot review on this slice + the post-merge passes on #167/#168 + the in-flight #174, the same area has now produced ≥12 findings:

- RFC 8941 SF-parameter handling (silent rejection / strip-then-check)
- Multi-algorithm Content-Digest entry ordering vs \`requireDigest\`
- \`requireDigest\` case sensitivity
- \`requireDigest=[""]\` Pydantic-vs-schema gap
- Schema-vs-Pydantic drift (twice — once in #173, again in #174)
- Parser fallback semantics + caller-site rejection logic split
- \`meta.digest\` algorithm coupling
- Rehash branch coverage (twice)
- Per-chunk \`asyncio.to_thread\` (rejected as a symmetry concern, but the question itself recurred)
- Hardcoded \`hashlib.new("sha256")\` vs the centralised map
- Empty-hex / no-colon / label-strip edges in the extracted verifier

That cluster size is what CLAUDE.md's TDD-discipline section calls *"common root cause across multiple defects → stop, return to design"*. Each finding has been mechanically addressable, but the iteration is converging on the same area: a hand-rolled RFC 8941 / 9530 parser layered with \`requireDigest\` policy and a schema-Pydantic-route triple source of truth.

Marking this PR back to **draft** while the digest pipeline is restructured. Two follow-ups identified:

1. **Replace the hand-rolled SF parser** — likely with [\`http_sfv\`](https://pypi.org/project/http-sfv/) or equivalent — so RFC 8941 parsing is a library concern rather than a per-edge maintenance burden.
2. **Decouple parse / verify-policy / requireDigest enforcement** into testable layers, and re-tighten the schema↔Pydantic agreement (memory observation 3689 from May 21 flagged this "triple source of truth" pattern back then; we're hitting its consequences now).

#174 will land first (the narrow \`[""]\` rejection fix + the schema mirror it implies) since that closes the immediate wire-validation gap. The bigger restructure will follow, then #169 rebases on top.

This is the "stop and design" the abandon-rule prescribes — applying findings 1-by-1 in another round is the whack-a-mole this discipline exists to prevent.

pvliesdonk:
## Restructure complete; flipping back to ready

The digest-pipeline restructure landed across two PRs as planned:

- **PR-A** ([#175](https://github.com/pvliesdonk/fastmcp-pvl-core/pull/175), **merged**) — added \`http-sf\` dependency + new \`_content_digest.py\` module wrapping the library + a contract-test suite covering every prior route-level edge.
- **PR-B** (this branch, rebased on PR-A's main) — three commits:
  - \`5a80dd3\`: moved \`_rehash_file\` to \`_staging.py\` (chunk-read primitive alongside \`_write_chunk\`)
  - \`5001439\`: deleted the hand-rolled \`_content_digest_parse\` / \`_content_digest_format\` from \`_upload.py\`; route now orchestrates \`_content_digest.parse_header\` + \`_content_digest.satisfies_requirement\` + the inline verify-against-bytes branch
  - \`60a2a5a\`: trimmed \`test_upload_helpers.py\` to media-range only (Content-Digest tests moved to \`test_content_digest.py\` in PR-A)

Every prior route test still passes — the migration is behaviour-preserving. The matrix's row count for slice 3 is unchanged. The Content-Digest edge surface that produced ≥12 findings over three review rounds is now contract-tested at the module level, with the route's role narrowed to orchestration.

The wire-spec gap that surfaced as PR #174 (\`requireDigest=[""]\`) is tracked separately at [\`pvliesdonk/mcp-file-exchange-ext#16\`](https://github.com/pvliesdonk/mcp-file-exchange-ext/issues/16); pvl-core does not fork the vendored spec.

633 file-exchange tests pass on Python 3.10 + 3.13; ruff + mypy clean.

pvliesdonk:
## Response to post-restructure review on 0e50fc9

Two findings applied as tests; one informational rejected with substance.

### Applied

| # | Severity | Fix |
|---|---|---|
| 1 | CONFIRMED gap | New \`test_route_consume_raises_still_returns_204\` patches \`store.consume\` to raise \`RuntimeError\` after a successful sink store; asserts the response is still 204 and the sink received the bytes. Covers the \`except Exception\` branch around \`consume()\` that was previously only exercised on its consumed=False path. |
| 2 | MAINTENANCE | New \`test_supported_algorithms_match_staging_hashlib_map\` pins the invariant that \`_content_digest.SUPPORTED_ALGORITHMS\` and \`_staging._HASHLIB_BY_LABEL\` have identical key sets. A future algorithm addition on only one side now fails loudly at test time instead of \`KeyError\` at runtime in the rehash branch. |

### Rejected: Finding 3 — \`rec.metadata["artifact_id"]\` not wrapped

The finding flagged that \`artifact_id\` dict access has no try/except, while the adjacent \`ArtifactConstraints.model_validate\` does. The asymmetry is real but justified — they cover different threat models:

- \`model_validate\` is wrapped because the rich \`expected\` object is exactly where schema-version drift between mint-side and route-side would surface (a future spec adds a field; mint writes new shape; old route reads new shape via Pydantic and gets a real \`ValidationError\` describing the drift). That's a structured failure mode worth a structured response.
- \`artifact_id\` is a flat string written by pvl-core's own mint side at \`upload_receiver_mint\`. There is no schema-evolution path where \`artifact_id\` becomes a non-string or disappears. A \`KeyError\` here would mean someone wrote completely different metadata under this token's key — a KV corruption mode the route has no useful response to.

Per CLAUDE.md: *"Don't add error handling, fallbacks, or validation for scenarios that can't happen. Trust internal code and framework guarantees."* The mint→route metadata contract is internal; the only field on it that carries plausible drift risk is \`expected\`, which is already wrapped.

If a future change adds richer fields to the mint-side metadata that genuinely could drift, the right move is wrapping that specific field — not a blanket \`try: dict access\`.

### CI

635 file-exchange tests pass on 3.10 + 3.13 (was 633, +2 new); ruff + mypy clean. PR #170 and #171 rebased on top.

PR #170, opened 2026-05-28
feat(file-exchange): cross-transport register_file_exchange_routes (slice 4/5 of #146)

pvliesdonk:
## Summary

Fourth of 5 stacked PRs for #146. Builds on #169 (upload route). Moves the public name \`register_file_exchange_routes\` out of \`_download.py\` into a new \`_routes.py\` and adds the upload-route arm.

- New \`_routes.py\` validates ALL preconditions (\`source\`-or-\`sink\`, sink-needs-\`config\`) BEFORE mounting any route (matrix row A7), then delegates to per-transport registrars.
- \`_download.py\`'s internal helper renamed to \`register_download_route\`.
- Public name \`register_file_exchange_routes\` is unchanged in \`__init__.py\` and \`file_exchange.py\` — just re-exported from \`_routes\` instead.
- New \`upload_receiver_mint\` re-export added to both public namespaces (alphabetical).

## Test plan

- [x] \`uv run pytest tests/_file_exchange\` — 619 passed (= 613 from slice 3 + 6 new registrar tests)
- [x] \`uv run --python 3.10 pytest tests/_file_exchange\` — 619 passed
- [x] \`uv run --python 3.13 pytest tests/_file_exchange\` — 619 passed

**Base:** depends on #169 → #168 → #167.

Part of #146.

### DIFF
--- a/src/fastmcp_pvl_core/_file_exchange/__init__.py
+++ b/src/fastmcp_pvl_core/_file_exchange/__init__.py
@@ -19,7 +19,6 @@ from fastmcp_pvl_core._file_exchange._codes import (
 from fastmcp_pvl_core._file_exchange._download import (
     download_fetcher_consume,
     download_provider_mint,
-    register_file_exchange_routes,
 )
 from fastmcp_pvl_core._file_exchange._errors import (
     FileExchangeTransferError,
@@ -44,6 +43,7 @@ from fastmcp_pvl_core._file_exchange._paths import (
     load_volume_map,
     resolve_filesystem_uri,
 )
+from fastmcp_pvl_core._file_exchange._routes import register_file_exchange_routes
 from fastmcp_pvl_core._file_exchange._selection import (
     select_sink,
     select_source,
@@ -67,6 +67,7 @@ from fastmcp_pvl_core._file_exchange._tokens import (
     build_capability_token_store,
     capability_url,
 )
+from fastmcp_pvl_core._file_exchange._upload import upload_receiver_mint
 from fastmcp_pvl_core._file_exchange._validation import (
     WireFormatError,
     validate_wire,
@@ -140,5 +141,6 @@ __all__ = [
     "resolve_filesystem_uri",
     "select_sink",
     "select_source",
+    "upload_receiver_mint",
     "validate_wire",
 ]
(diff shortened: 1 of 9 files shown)


pvliesdonk:
## Response to bot review on 33206fb

All six findings applied, including the CONFIRMED bug. Two new regression tests added.

### Applied

| # | Source | Severity | Fix |
|---|---|---|---|
| 1 | Claude | **CONFIRMED bug** | \`_upload.py\` route: \`rec.metadata["artifact_id"]\` now uses \`.get()\` + returns 404 when None. This is the **first PR to mount both data planes on a unified \`token_store\`**, so a download-minted token (metadata \`{"key": …}\`) could reach the upload route and trigger a KeyError as an unstructured ASGI 500. My earlier rejection of the analogous \`artifact_id\` finding on PR #169 (where I argued "mint side controls writes") was correct in isolation but wrong in slice 4's context — a real root-cause-becomes-visible-only-when-a-later-slice-lands pattern. |
| 2 | Claude | Minor | \`_download.py\` route: same fix symmetrically for \`rec.metadata["key"]\`. The bare subscript was inside try/except so it didn't crash, but the \`logger.exception\` message would have misattributed the failure to the source hook. Returns 404 before opening the source now. |
| 3 | Claude | Minor | Replaced \`cast("ServerConfig", config)\` in \`_routes.py\` with \`assert config is not None\` + comment noting the precondition gate is the load-bearing guarantee. |
| 4 | Claude | Test symmetry | \`test_registrar_both_none_raises_value_error\` now also asserts no \`/fx/\` routes mounted — matches the sibling \`test_registrar_precondition_failure_mounts_nothing\`. |
| 5 | Claude | Test gap | Added \`test_upload_data_plane_names_reexported\` to the namespace test mirroring the existing download-data-plane test. |
| 6 | Gemini | CLAUDE.md compliance | \`register_file_exchange_routes\` docstring now documents each kwarg's CLAUDE.md category (shape / hook / config). |

### Two new regression tests for the cross-transport-token scenario

- \`test_cross_transport_token_upload_route_returns_404\` — mints a download token, PUTs it at \`/fx/u/{token}\`, asserts 404.
- \`test_cross_transport_token_download_route_returns_404\` — mirror: mints an upload token, GETs it at \`/fx/d/{token}\`, asserts 404.

These pin the cross-transport contract so a future regression that re-introduces the bare subscript fails loudly.

### CI

650 file-exchange tests + 10 namespace tests = 660 total pass on Python 3.10 + 3.13; ruff + mypy clean. PR #171 rebased on top of these fixes.

PR #174, opened 2026-05-28
fix(file-exchange): reject empty-string elements in requireDigest

pvliesdonk:
## Summary

PR #169 (slice 3) surfaced that \`ArtifactConstraints.requireDigest\`'s field-level constraint validates only the list length, not each algorithm name. A mint with \`requireDigest=[\"\"]\` passes validation and creates a token that returns 400 on every upload forever (the route's lowercased \`preferred_set={\"\"}\` matches no real algorithm label, and the fallback then fails the requireDigest membership check).

Fix: add an inner \`Annotated[str, Field(min_length=1)]\` so each element must be a non-empty string.

## Test plan

- [x] \`uv run pytest tests/_file_exchange\` — 595 passed (594 + 1 new)
- [x] \`uv run --python 3.10 pytest tests/_file_exchange\` — 595 passed
- [x] \`uv run --python 3.13 pytest tests/_file_exchange\` — 595 passed
- [x] \`uv run ruff format --check .\` clean

## Why a separate PR

Same logic as #173: small fix to a wire-level function the in-flight #146 chain (#169 / #170 / #171) depends on. Landing it on its own keeps slice 3's review surface narrow; #169 will inherit on rebase.

Refs #146.

### DIFF
--- a/src/fastmcp_pvl_core/_file_exchange/_wire.py
+++ b/src/fastmcp_pvl_core/_file_exchange/_wire.py
@@ -121,7 +121,15 @@ class ArtifactConstraints(_WireBase):
 
     maxSize: int | None = Field(default=None, ge=0)  # noqa: N815
     acceptMimeTypes: list[str] | None = None  # noqa: N815
-    requireDigest: list[str] | None = Field(default=None, min_length=1)  # noqa: N815
+    # ``min_length=1`` on the outer ``Field`` constrains the LIST length
+    # (≥1 algorithm), while the inner ``Annotated[str, Field(min_length=1)]``
+    # constrains each algorithm name to be a non-empty string. Without the
+    # inner constraint, ``requireDigest=[""]`` would pass validation and
+    # then permanently 400 every upload against that token (the route's
+    # ``preferred_set={""}`` matches no real algorithm).
+    requireDigest: list[Annotated[str, Field(min_length=1)]] | None = Field(  # noqa: N815
+        default=None, min_length=1
+    )
 
 
 class FilesystemSource(_DescriptorBase):
--- a/tests/_file_exchange/test_wire_metadata.py
+++ b/tests/_file_exchange/test_wire_metadata.py
@@ -59,6 +59,14 @@ def test_constraints_require_digest_must_be_non_empty():
         ArtifactConstraints(requireDigest=[])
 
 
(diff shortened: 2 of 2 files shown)


pvliesdonk:
## Closing — this PR forks the vendored wire spec

Per CLAUDE.md:

> *"Real spec gaps are resolved through a proper spec evolution: a new release with the version field bumped per the spec's own versioning-and-compatibility section. **Inline amendments to a published version are not a valid spec-evolution mechanism.**"*
>
> *"If pvl-core's implementation is wrong, or diverges from a \`docs/specs/\` spec, the resolution is to fix pvl-core centrally — one change, every downstream follows — or to evolve the spec."*

\`src/fastmcp_pvl_core/_file_exchange/_schema/file-exchange.json\` is a vendored copy of the [\`mcp-file-exchange-ext\` v0.1 spec](https://github.com/pvliesdonk/mcp-file-exchange-ext) pinned to commit \`5f50a4e…\` (per EPIC #138). Tightening pvl-core's Pydantic model to reject \`requireDigest=[""]\` makes pvl-core's runtime stricter than the schema — exactly the fork-the-spec behaviour CLAUDE.md prohibits.

Both bot reviewers (Claude HIGH, Gemini HIGH) **independently surfaced this**: their requested "update the schema to match Pydantic" fix is correct as a diagnosis but not a valid in-PR fix here, because pvl-core does not own that schema.

### What replaces this PR

1. The route-level guard already in slice 3 handles \`requireDigest=[""]\` correctly: after lowercasing, \`cd_algo not in {""}\` evaluates to True and the upload returns 400. Tokens minted with empty-string algorithm names are unusable, but that's the downstream's programming-error symptom, not a security gap.
2. Filed [pvliesdonk/mcp-file-exchange-ext#...](TBD) proposing \`minLength: 1\` on \`requireDigest\` items as a v0.2 spec evolution. When that lands, pvl-core re-pins and the constraint comes through the vendored schema correctly.

This is part of a larger root-cause pause on the digest pipeline (see [PR #169](https://github.com/pvliesdonk/fastmcp-pvl-core/pull/169#issuecomment-4563641353)) — the cluster of bot findings around Content-Digest / requireDigest signals over-validation drift, hand-rolled SF parsing, and the schema-Pydantic-route triple source of truth. Whack-a-mole on individual findings was the wrong response.

PR #175, opened 2026-05-28
feat(file-exchange): _content_digest.py + http-sf dependency (#146)

pvliesdonk:
## Summary

Root-cause response to the ≥12-finding cluster on PRs #167 / #168 / #169 / #173 / #174 (closed). Introduces a new pure-function module \`_content_digest.py\` wrapping the [\`http-sf\`](https://pypi.org/project/http-sf/) library (Mark Nottingham's RFC 8941 reference implementation, pure-Python) plus a comprehensive contract-test suite covering every Content-Digest edge previously surfaced as a route-level bug.

The route itself is untouched. **PR-B** (rebased slice 3 / #169) will delete the hand-rolled parser and migrate the route to call into this module.

## Design

\`docs/superpowers/specs/2026-05-28-content-digest-restructure.md\` — 4-layer architecture (library handles RFC 8941 grammar; \`_content_digest.py\` owns algorithm selection + requireDigest membership; the route still owns bytes verification because that's tied to the staging temp-file lifecycle).

\`docs/superpowers/plans/2026-05-28-content-digest-restructure.md\` — implementation plan (PR-A as 7 tasks; PR-B as 5).

## Surface added

- \`_content_digest.SUPPORTED_ALGORITHMS\` — frozenset of algorithm labels pvl-core supports (\`sha-256\`, \`sha-384\`, \`sha-512\`).
- \`_content_digest.parse_header(header, *, preferred=None) -> (algo, bytes) | None\` — RFC 9530 §3 MUST-ignore semantics; honours \`preferred=\`; falls back to first supported entry.
- \`_content_digest.satisfies_requirement(algo, required) -> bool\` — case-insensitive membership against \`ArtifactConstraints.requireDigest\`.
- \`_content_digest.format_header(algo, raw) -> str\` — RFC 9530 serialiser via \`http_sf.ser_dictionary\`.

## Edge surface pinned by tests

Every prior route-level finding now has a contract test at the module level:

| Edge | Test |
|---|---|
| Empty / malformed header | \`test_parse_header_empty_returns_none\`, \`..._malformed_returns_none\` |
| All-unsupported algos | \`..._all_unsupported_returns_none\` |
| RFC 9530 §3 MUST-ignore | \`..._skips_unsupported_takes_supported\` |
| RFC 8941 §3.2 params on entries | \`..._ignores_sf_parameters\` |
| Uppercase algorithm label | \`..._uppercase_label_rejected\` (http_sf rejects per lcalpha grammar) |
[...]

PR #176, opened 2026-05-29
feat(file-exchange): #148 umbrella helpers + Tasks integration + adoption docs

### LINKED ISSUE #148: file-exchange (10/10): top-level register_file_exchange_* helpers + Tasks integration + adoption docs

pvliesdonk:
## Summary

Final EPIC #138 child task. Adds the opinionated tool-registration layer on top of the data-plane primitives merged across #167 / #168 / #169 / #170 / #171 / #173 / #175.

- \`register_file_exchange(mcp, *, config, base_url, source=None, sink=None, volume_map=None) -> FileExchangeContext\` — setup call: builds the token store, mounts the routes via the existing \`register_file_exchange_routes\`, declares the server-level Tasks capability.
- \`@register_file_exchange_provider(mcp, tool_name, fxctx)\` — decorator: downstream returns \`(metadata, key)\`; the wrapper mints a \`TransferHandle\`.
- \`@register_file_exchange_receiver(mcp, tool_name, fxctx)\` — decorator: downstream returns \`(artifact_id, expected)\`; the wrapper mints an \`IntakeTicket\`.
- \`register_file_exchange_fetcher(mcp, tool_name, fxctx)\` — generates a tool that takes a \`TransferHandle\`, selects a source descriptor, dispatches to the right transport's fetcher.
- \`register_file_exchange_sender(mcp, tool_name, fxctx)\` — generates a tool that takes an \`IntakeTicket\` + a local key, selects a sink descriptor, dispatches to the right transport's sender.
- Every helper-registered tool carries \`taskSupport=\"optional\"\` per §14 (verified empirically against \`mcp.tool(annotations=...)\` ⇒ \`ToolAnnotations\` carries the unknown key through to the wire form because of \`additionalProperties=True\`).
- \`tasks.requests.tools.call\` declared server-level via \`mcp.experimental_capabilities\` (chosen over \`FastMCP(tasks=True)\` to avoid pulling in the docket extra — we declare the *capability*, not the scheduler binding).

## Design correction from the plan

The original plan + design called \`filesystem_fetcher_consume(..., config=...)\` — but the actual signature is \`(..., volume_map: VolumeMap)\`. \`FileExchangeContext\` gained \`volume_map: VolumeMap | None = None\` so the filesystem transport is opt-in. A fetcher/sender tool that selects a filesystem descriptor without a \`volume_map\` raises \`FileExchangeTransferError(NO_SUPPORTED_TRANSPORT)\` at call time. The download/upload transports work without a volume_map.

## Documentation

- \`README.md\` — \`## File-exchange extension\` section.
- \`docs/file-exchange.md\` — pvl-core implementation notes (architecture, setup walkthrough, Tasks integration, filesystem opt-in, error model).
- \`docs/file-exchange-adoption.md\` — one worked example per role + filesystem opt-in + multi-role note.
- \`CHANGELOG.md\` — entry under the existing \`## [3.0.0] - UNRELEASED\` section.

## Design + plan

- \`docs/superpowers/specs/2026-05-28-file-exchange-148-umbrella-helpers-design.md\`
- \`docs/superpowers/plans/2026-05-28-file-exchange-148-umbrella-helpers.md\`

## Test plan

- [x] Per-role unit tests in \`tests/_file_exchange/test_helpers.py\` (24 tests).
- [x] Two-server e2e tests in \`tests/_file_exchange/test_helpers_e2e.py\` (provider→fetcher and receiver→sender).
- [x] Namespace test extended to cover the new public surface.
- [x] \`uv run pytest tests/_file_exchange tests/test_file_exchange_namespace.py\` — 696 passed.
- [x] \`uv run --python 3.10 pytest …\` — 696 passed.
- [x] \`uv run --python 3.13 pytest …\` — 696 passed.
- [x] \`uv run ruff format --check . && uv run ruff check src tests\` clean.
- [x] \`uv run mypy src\` — no issues in 36 source files.

Closes #148.


pvliesdonk:
## Response to bot review on 1696147

Three Claude findings applied; four Gemini findings rejected with empirical / spec-discipline reasoning.

### Applied (Claude × 3)

| # | Severity | Fix |
|---|---|---|
| 1 | CONFIRMED | Removed dead \`if fxctx.volume_map is None\` guards inside the \`isinstance(FilesystemSource/Sink)\` arms of both fetcher and sender. The branches were unreachable because \`select_source\`/\`select_sink\` with \`is_accessible=None\` already skip every filesystem descriptor (per \`_selection.py\` docstring: *"None means the party does not support filesystem at all — every filesystem source is skipped."*). The two existing \`*_filesystem_without_volume_map_*\` tests still pass — they hit the outer \`descriptor is None\` branch now (correct behaviour, same §13 envelope). |
| 2 | PLAUSIBLE | \`mcp.experimental_capabilities["tasks"]\` switched from blind \`=\` to a \`setdefault\` chain so a future FastMCP that pre-populates other sub-keys isn't silently clobbered. |
| 3 | PLAUSIBLE | Added docstrings to the generated \`_consume_transfer\` and \`_send_to_receiver\`. FastMCP uses the function docstring as the MCP \`description\` field in \`tools/list\`; without one, peers saw blank descriptions. |

### Rejected with substance (Gemini × 4)

**Gemini HIGH × 2** — *replace \`is_accessible=None\` with \`lambda _: False\`*

Empirically verified the opposite of Gemini's claim:

\`\`\`python
>>> from fastmcp_pvl_core._file_exchange._selection import select_source
>>> handle = TransferHandle(..., sources=[FilesystemSource(...), DownloadSource(...)])
>>> select_source(handle, is_accessible=None)
DownloadSource(...)   # filesystem skipped, fell through to download — as documented
\`\`\`

\`_selection.py\`'s docstring is explicit: *"\`None\` means the party does not support filesystem at all — every filesystem source is skipped."* Gemini's \`lambda _: False\` suggestion is a no-op rewrite of \`None\`. The Claude-CONFIRMED dead-guard finding above is the correct diagnosis of the same code; both reviewers cannot be right and the empirical evidence sides with Claude.

**Gemini MEDIUM × 2** — *add \`inspect.iscoroutinefunction\` for sync callback support*

The plan (\`docs/superpowers/plans/2026-05-28-file-exchange-148-umbrella-helpers.md\`) documents the provider/receiver decorators as async-only — the spec's tool registration path uses \`async def\` bodies throughout, the e2e tests confirm this, and FastMCP itself supports sync tools via a different code path that doesn't go through these wrappers. Adding sync-callback support is feature creep beyond the task scope.

Per CLAUDE.md: *"Don't add features beyond what the task requires."*

If a future downstream genuinely needs sync provider/receiver bodies, it's a clean follow-up — the existing async-only contract is documented and tested. Adding the \`iscoroutinefunction\` dispatch reflexively here would mean shipping untested code paths.

### CI

696 file-exchange + namespace tests pass on Python 3.10 + 3.13; ruff + mypy clean.

pvliesdonk:
@claude review

pvliesdonk:
Round 2 + round 3 findings applied in 4242922.

**R2-1 — wire inputSchema coverage for provider/receiver decorators.** Added `test_provider_decorator_preserves_inner_signature_in_wire_schema` and the receiver mirror. Each asserts the inner-fn parameter (`report_id` / `case_id`) is present in `tool.to_mcp_tool().inputSchema["properties"]` — i.e. the schema MCP clients actually see, not just the Python callable.

**R2-2 — anchor `transport=None` on the no-volume-map error path.** Added `assert ei.value.transport is None` to both `test_fetcher_filesystem_without_volume_map_raises_transfer_error` and the sender mirror. The error fires from the outer `descriptor is None` branch (filesystem skipped at selection because `volume_map is None`), so `transport` is `None`, not `"filesystem"`. Pinned against a future refactor that incorrectly populates the transport field on that branch.

**R2-3 — runtime guard for the 2-tuple return contract.** Both decorators now raise `TypeError` with a message naming the helper and the tool_name if the user's inner function returns something other than a 2-tuple. Two new tests assert the type and the message content.

**R3-A — `functools.wraps` at module top.** Removed the lazy `from functools import wraps` from both `_wrap` closures and added the import to the module header.

**R3-B — strip "Generated by …" from generated tool docstrings.** `_consume_transfer` and `_send_to_receiver` no longer mention the internal helper name in their docstrings, so it does not leak into MCP `tools/list` output.

Verified: `ruff format/check` clean, `mypy src` clean, `pytest` 1091 passed / 2 skipped on Python 3.10, and 683 passed on the `_file_exchange` slice under Python 3.13.

pvliesdonk:
Round-4 findings applied in 48ed5f4. Discipline acknowledgement first, then the changes.

**Discipline slip — root cause.** Last round I claimed R2-2 was applied to "both the fetcher and the sender mirror." It was applied to the fetcher and to the **wrong** sender test: I located the sender edit site by grepping for the surrounding code pattern (`await tool.fn(ticket=ticket, key="k")` followed by the `code ==` assert) without verifying which `async def test_…` actually enclosed it. That pattern matches the body of both `test_sender_filesystem_without_volume_map_raises_transfer_error` and `test_sender_no_supported_transport_raises_transfer_error`; I edited the second when I meant the first, and the announce-message read "both" because I'd reasoned about the intent not the diff.

The right discipline here per [`feedback_sweep_repo_wide`](memory:feedback_sweep_repo_wide) is: when fixing one instance of a repo-wide class of finding, grep for *every* instance up front and sweep them in one pass. I should have run `grep -n "NO_SUPPORTED_TRANSPORT" tests/_file_exchange/test_helpers.py` last round (it would have shown four sites — two anchored, two unanchored) instead of jumping to the assumed two.

**Applied in this round (one sweep):**

- **Finding 1 (CONFIRMED miss).** `test_sender_filesystem_without_volume_map_raises_transfer_error` line 653 — added `assert ei.value.transport is None`.
- **Finding 2 (PLAUSIBLE).** `test_fetcher_no_supported_transport_raises_transfer_error` line 726 — added `assert ei.value.transport is None`. `select_source` skips `UnknownTransportDescriptor` via `continue` (`_selection.py` lines 87–88) and returns None, so the outer `descriptor is None` branch fires with `transport=None`, matching the sender mirror.
- **Finding 3 (COSMETIC).** `register_file_exchange` docstring — replaced "(in a later task)" with the present-tense description; the Tasks capability declaration is implemented in this function via the `setdefault` chain.

All four `NO_SUPPORTED_TRANSPORT` sites are now anchored with `transport is None`. Verified: ruff format/check clean, mypy clean, 683 tests pass in the file-exchange slice.

Issue #181, opened 2026-06-03
refactor: converge _debug.py DEBUG_PORT parse onto env_int (soft mode)


pvliesdonk:
## Context

Follow-up from #180, which adds the public dual-mode `env_int` / `env_float` helpers and converts `ServerConfig.from_env`'s `port` parse to use `env_int(..., strict=True, minimum=1, maximum=65535)`.

`start_debugpy_if_configured` in `src/fastmcp_pvl_core/_debug.py` is the **second internal consumer** of the exact soft-mode shape `env_int` now provides: it hand-rolls `int()` + `1..65535` range check + `WARNING`-and-skip on a bad/out-of-range `{PREFIX}_DEBUG_PORT`. It was deliberately left out of #180 to keep that PR's diff focused and avoid churning the `test_debug.py` hotspot under a feature PR.

## Why it isn't a drop-in

`_debug.py` carries a **domain sentinel `env_int` does not model**: `DEBUG_PORT == 0` (and `00`, `-0`, `+0`, whitespace-wrapped) is the *documented disable form* — a **silent** no-op, not a warning. A naive `env_int(minimum=1, ...)` would emit a WARNING on `0`, changing documented behavior.

## Proposed change

Rewrite the parse as:

```python
port = env_int(env_prefix, "DEBUG_PORT", None, minimum=0, maximum=65535)
if port is None or port == 0:
    return  # unset/blank/invalid (already warned by env_int) or documented disable form
```

- `minimum=0` so `0` is *valid* (returned), then `_debug` treats `0`/`None` as the disable form — preserving the silent-on-`0` contract.
- Negatives / non-numeric / `> 65535` → `env_int` soft-mode WARNING + `None` → listener not started, matching today's outcome.

## Behavior delta to verify

- **Warning message text changes** from the bespoke `"%s=... debugpy listener not started."` to `env_int`'s generic `"<KEY> must be an integer; got <raw> — using default None"` / `"<KEY> must be <= 65535; got <n> — using default None"`.
- `test_debug.py` warning-message assertions are updated to the new text (they currently assert the key name + WARNING level + no double-underscore; confirm each still holds and update the bespoke-phrasing assertions). `test_debug.py` is a churn hotspot — touch it deliberately.

## Deliverables

- [ ] `_debug.py` consumes `env_int`; the bespoke `try/except int()` + range block is **gone** (grep: no `int(raw)` / `1 <= port <= 65535` left in `_debug.py`).
[...]

PR #182, opened 2026-06-03
feat(env): public dual-mode numeric env helpers (env_int/env_float)

### LINKED ISSUE #159: Harden ServerConfig.from_env numeric env parsing (clear errors + value validation)
## Problem

`ServerConfig.from_env` parses numeric environment variables with bare
`int()` / `float()` calls and no value-range validation:

- `src/fastmcp_pvl_core/_config.py:146` — `port=int(port_str)`
- `src/fastmcp_pvl_core/_config.py:160` — `file_exchange_token_ttl=float(token_ttl_str)`
- `src/fastmcp_pvl_core/_config.py:162` — `file_exchange_max_artifact_size=(int(max_size_raw) if max_size_raw else None)`

Two gaps, both pre-existing (the `port` parse predates #144; the two
file-exchange fields followed the same established pattern in #144):

1. **Unhelpful error on malformed input.** A bad value (e.g.
   `MYAPP_FILE_EXCHANGE_TOKEN_TTL=abc`) raises a raw
   `ValueError: could not convert string to float: 'abc'` that does not
   name the offending env var. The operator gets a stack trace pointing
   at `_config.py`, not "your `MYAPP_FILE_EXCHANGE_TOKEN_TTL` is invalid."
   `ConfigurationError` exists precisely for operator-actionable,
   fail-fast-at-load misconfiguration.

2. **No value-range validation at the config layer.** A non-positive
   `file_exchange_token_ttl` (e.g. `0` / `-5`) parses fine and only fails
   later, at `CapabilityTokenStore` construction (`ttl_ceiling must be
   positive`), detached from the config-load site. A negative
   `file_exchange_max_artifact_size` (e.g. `-1`) is accepted silently and
   would be a footgun for the #146 upload route that consumes it.

This surfaced repeatedly as a sub-80 (non-blocking) observation from
multiple local-review lenses during the #144 review — correct behavior
(it does fail loud, no silent coercion), just poor message quality and
late/absent range checks. It is deliberately **out of #144's scope**
because the fix is a repo-wide sweep of all `from_env` numeric parses,
not a #144-local change.

## Proposed change

Sweep all numeric `from_env` parses in one pass (per the "fix the class
repo-wide, not incrementally" discipline):

- Wrap each parse so a malformed value raises `ConfigurationError`
  naming the env var and showing the offending value, e.g.
  `ConfigurationError(f"{prefix}_FILE_EXCHANGE_TOKEN_TTL must be a number (seconds); got {value!r}")`.
- Reject semantically-invalid values at parse time: `port` in range,
  `file_exchange_token_ttl > 0`, `file_exchange_max_artifact_size >= 0`
  (or `> 0`) — fail at config load, not at downstream consumption.
- A small `_parse_int_env` / `_parse_float_env` helper keeps the three
  call sites DRY.

## Deliverables

- Apply to all three current numeric fields (`port`,
  `file_exchange_token_ttl`, `file_exchange_max_artifact_size`).
- Tests for: malformed value -> `ConfigurationError` naming the var;
  out-of-range value -> `ConfigurationError`; valid value -> parsed.
  Update existing `from_env` tests if they assumed bare `ValueError`.

### LINKED ISSUE #180: feat: public dual-mode numeric env helper for downstream reuse (env_int/env_float)
## Summary

`fastmcp_pvl_core` exposes `env`, `parse_bool`, `parse_list`, and `parse_scopes`, but **no public numeric (int/float) env-parsing helper**. Downstream projects therefore hand-roll numeric env parsing per field.

This is distinct from #159 / #161: those harden `ServerConfig.from_env`'s *own* internal numeric casts and propose a *private*, *raise-only* `_parse_int_env` / `_parse_float_env`. This issue is about a **public, exported, dual-mode** helper that downstreams can consume.

## Observed  [verified: read the source 2026-06-03]

- `src/fastmcp_pvl_core/_env.py` provides `env(prefix, name, default)` plus `parse_bool` / `parse_list` / `parse_scopes` only — no numeric parser. All four are exported from `__init__.py`; no numeric equivalent is.
- `markdown-vault-mcp`'s `config.py` `load_config()` hand-rolls numeric env parsing **nine times in two distinct shapes**:
  - *warn-and-fall-back-to-default* (try/except → `logger.warning` → default), 5×: `GIT_PUSH_DELAY_S`, `GIT_PULL_INTERVAL_S`, `FILE_WATCHER_DEBOUNCE_S`, `MAX_ATTACHMENT_SIZE_MB`, `MAX_NOTE_READ_BYTES`.
  - *parse-or-raise* (try/except → `raise ValueError`), 4×: `CHUNKS_PER_FILE`, `SNIPPET_WORDS`, `LENGTH_DOWNWEIGHT_ALPHA`, `MAX_CHUNK_WORDS`.
  - Most stanzas also apply a post-parse bound (e.g. `>= 1`, `>= 0`, or negative → default).
- Sister downstreams: `image-generation-mcp/config.py` parses one numeric env var (`int(...)` for `transform_cache_size`); `kroki-mcp`, `scholar-mcp`, `paperless-mcp` parse none. So MV is the heavy consumer; cross-repo reuse beyond MV is one callsite today.

## Expected / desired

A public, exported numeric env-parsing capability in `fastmcp_pvl_core`, alongside the existing `parse_*` family, that supports **both** observed shapes — warn-and-default for soft limits and raise for hard parameters — so downstreams read int/float env vars (with a default, and optionally a validity bound) without re-implementing try/except-and-fallback per field. Ideally `ServerConfig.from_env` (and the #159/#161 hardening) would consume the same helper internally.

## Context

- Related: #159 (harden `from_env` numeric parses → `ConfigurationError` + range validation; proposes a private `_parse_int_env`/`_parse_float_env`) and #161 (same theme + `http_timeout`; item 1 overlaps #159).
- Motivated by `markdown-vault-mcp` #579 (CollectionConfig sub-config restructure). That project's final PR (the `from_env` convergence) intends to consume this helper and is **blocked** on it landing + a pvl-core release + dependency bump. The earlier #579 PRs do not depend on this.

## Scope boundary

- NOT the internal `from_env` numeric hardening — that is #159/#161. (If #159's helper is built public + dual-mode, it may satisfy this issue; tracking separately per maintainer decision.)
- NOT changing the existing `env` / `parse_bool` / `parse_list` / `parse_scopes` helpers.
- Downstream adoption (replacing MV's nine stanzas, the one image-gen callsite) is out of scope here — it happens per-repo after release.

## Open questions  (design decisions for the implementer; all `[unverified]`)

- `[unverified]` Shape: a string→number parser mirroring `parse_bool`/`parse_list` (`parse_int(value)` / `parse_float(value)`), or a read-and-parse helper mirroring `env` (`env_int(prefix, name, default, ...)` / `env_float(...)`), or both. The module already uses both shapes (`env` reads; `parse_*` convert), so either is consistent.
- `[unverified]` Dual-mode expression: a `default` plus an opt-in `strict`/raise flag, vs separate functions, vs a sentinel for "no default → raise".
- `[unverified]` Bound validation: fold in optional `minimum`/`maximum` (several MV stanzas clamp/reject out-of-range), or leave bounds to the caller.
- `[unverified]` Error type in raise-mode: `ValueError` vs `ConfigurationError`.
- `[unverified]` Whether to converge with #159's proposed private helper (single helper, public + dual-mode, used internally and exported).

I have not verified which signature the maintainer prefers — these are design decisions, not observed facts.

pvliesdonk:
## Summary

Adds `env_int` / `env_float` to the env-helper family — public, read-and-parse numeric env parsers downstream servers can consume instead of hand-rolling `try/except`-and-fallback per field. Both support the **two shapes** `markdown-vault-mcp` hand-rolls nine times today:

- **soft** (`strict=False`, default): on an invalid/out-of-range value, log a `WARNING` and return `default` — for soft limits.
- **strict** (`strict=True`): on an invalid/out-of-range value, raise `ConfigurationError` naming the var — for hard parameters.

Both take a `default` plus optional inclusive `minimum`/`maximum` bounds (reject-semantics, never clamp). `env_float` additionally rejects non-finite values (`nan`/`inf`/`-inf`). An unset/blank var returns `default` silently in both modes (delegates to `env()`'s strip semantics, keeping "unset" and "malformed" distinct).

`ServerConfig.from_env`'s `port` cast now consumes the helper: `env_int(env_prefix, "PORT", 8000, strict=True, minimum=1, maximum=65535)`.

## Behavior change (the live scope of #159)

`from_env` previously parsed `port` with a bare `int(port_str)`:
- a malformed `PORT` raised a context-free `ValueError` from `_config.py` → now raises `ConfigurationError` naming the var;
- `PORT` of `0` / `-5` / `70000` were **silently accepted** → now raise `ConfigurationError` (a server *listen* port outside `1..65535` is a misconfiguration).

## Closes

- **Closes #180** — the public dual-mode numeric helper itself.
- **Closes #159** — `env_int` supersedes #159's proposed *private* `_parse_int_env`/`_parse_float_env` with a public, dual-mode helper used internally for `port`. `port` is #159's only live numeric cast on `main`; its two file-exchange casts (`token_ttl`, `max_artifact_size`) no longer exist (parked off main), so this fully resolves #159's actionable scope.

## Out of scope (tracked)

- **#181** — converging `_debug.py`'s `DEBUG_PORT` (the second internal soft-mode callsite) onto `env_int`. Deferred because it carries a `port==0`-as-disable sentinel `env_int` does not model, and converting it churns the `test_debug.py` hotspot.
- Downstream adoption (MV's nine stanzas, image-gen's one callsite) — per-repo, after a pvl-core release.

## Design

`docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md` — semantics matrix, error-message format, and the full failure-mode/test enumeration.

## Tests

TDD, one behavior each. `tests/test_env.py` (`TestEnvInt`, `TestEnvFloat`): unset/blank/valid/malformed/out-of-range across soft+strict, inclusive boundaries, `env_int` rejects float text, `env_float` rejects non-finite, prefix-underscore + no-double-underscore-in-key. `tests/test_config.py`: `from_env` port unset→8000, malformed→`ConfigurationError`, `0`/`-1`/`70000`→raise, `1`/`65535`→accepted.

Local checks green on this branch: `uv run pytest` (444), `ruff format --check`, `ruff check`, `mypy src`.

### DIFF
--- a/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
+++ b/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
@@ -115,6 +115,26 @@ This is a deliberate **behavior change** — the live scope of #159:
 `port == 0` is rejected: a server *listen* port of 0 (OS-ephemeral) is a
 misconfiguration for an addressable MCP server.
 
+### Versioning (deliberate `feat:`/minor, not breaking)
+
+Although this tightens `from_env`'s PORT handling — malformed input now raises
+`ConfigurationError` (not a bare `ValueError`; note `ConfigurationError` is **not**
+a `ValueError` subclass) and `0`/negative/`>65535` now raise instead of being
+accepted — it ships as a non-breaking **minor** (`feat:`), not a major. Rationale,
+recorded so the choice is visible to review and not a missing-`BREAKING CHANGE`
+oversight:
+
+- The old behaviour was never a documented contract: the bare `int(port_str)`
+  `ValueError` was incidental, and out-of-range acceptance was the latent bug
+  #159 was filed to fix. Tightening validation of *invalid* input is the
+  hardening #159 explicitly asked for, not a redesign of a promised API.
+- The consumers are the coordinated `pvliesdonk/*-mcp` family; none binds a
+  non-`1..65535` PORT. No external contract relies on the old exception type.
+
+This is the maintainer's semver decision. A reviewer (or `python-semantic-release`)
+expecting a `BREAKING CHANGE` marker for an exception-type change should read this
+note as the deliberate, accepted answer rather than an omission.
+
 ### Exports
 
 `env_float`, `env_int` added to `__init__.py` import and `__all__` in alphabetical
--- a/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
+++ b/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
@@ -70,6 +70,20 @@ silently changing an operator's configured value is a footgun.
 - Soft mode logs the **same message text** at `WARNING` via a module logger
   (`logging.getLogger("fastmcp_pvl_core._env")`), suffixed `— using default <default>`.
 
+### Accept-set (shape decision, recorded during review)
+
+The helpers **delegate to Python's `int()` / `float()`** and accept whatever those
+accept — including PEP 515 underscore separators (`"1_000"` → `1000`), a leading
+sign, and (for `float` only) scientific notation (`"1e3"` → `1000.0`). This is a
+deliberate shape decision: rejecting underscores alone would be incoherent while
+still accepting `1e3`/`+5`/non-ASCII digits, and a true "plain decimal only"
+contract would need a regex for no real benefit (readable large numbers like
+`10_000_000` are a plus for downstream byte/size limits). The accept-set is
+**documented in both docstrings and pinned by characterization tests** so it is an
+explicit, regression-protected contract rather than an implicit consequence of the
+stdlib. `int()`/`float()` reject leading/trailing/doubled underscores themselves,
+so `"_1"` / `"1__0"` are still invalid.
+
 ### Small refactor
 
 Extract `_resolve_key(prefix, name) -> str` (`f"{prefix.rstrip('_')}_{name}"`) and use
--- a/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
+++ b/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
@@ -56,7 +56,10 @@ raise/warn only fires on malformed-when-set, never on a simply-absent var).
 
 `minimum` / `maximum` are inclusive and **reject** out-of-range values (treated
 identically to malformed: soft warns + default, strict raises). No clamping/coercion —
-silently changing an operator's configured value is a footgun.
+silently changing an operator's configured value is a footgun. The bounds validate the
+**operator's env value only**; the developer-supplied `default` is the trusted fallback
+and is returned as-is (never re-validated against the bounds), consistent with the
+unset path.
 
 ### Errors and messages
 
@@ -76,7 +79,8 @@ The helpers **delegate to Python's `int()` / `float()`** and accept whatever tho
 accept — including PEP 515 underscore separators (`"1_000"` → `1000`), a leading
 sign, and (for `float` only) scientific notation (`"1e3"` → `1000.0`). This is a
 deliberate shape decision: rejecting underscores alone would be incoherent while
-still accepting `1e3`/`+5`/non-ASCII digits, and a true "plain decimal only"
+still accepting `+5`/non-ASCII digits (and, for `float`, `1e3`), and a true "plain
+decimal only"
 contract would need a regex for no real benefit (readable large numbers like
 `10_000_000` are a plus for downstream byte/size limits). The accept-set is
 **documented in both docstrings and pinned by characterization tests** so it is an
--- a/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
+++ b/docs/superpowers/specs/2026-06-03-env-numeric-helper-design.md
@@ -1,0 +1,157 @@
+# Design: numeric env helpers (`env_int` / `env_float`)
+
+**Date:** 2026-06-03  
+**Issues:** pvliesdonk/fastmcp-pvl-core#180, #159 (Closes both — see closure note)  
+**Follow-up:** pvliesdonk/fastmcp-pvl-core#181 (`_debug.py` convergence — out of scope here)  
+**Related downstream:** pvliesdonk/markdown-vault-mcp#579 (blocked on this landing + a pvl-core release)
+
+---
+
+## Problem
+
+`fastmcp_pvl_core._env` exposes `env` (read) and `parse_bool` / `parse_list` /
+`parse_scopes` (convert) — but **no numeric parser**. Downstream servers therefore
+hand-roll int/float env parsing per field. `markdown-vault-mcp` does it **nine times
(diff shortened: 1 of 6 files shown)


pvliesdonk:
Addressed the review:

- **Missing `"Infinity"` in `test_non_finite_strict_raises`** — fixed in 82ad5ef. Added `"Infinity"` to the strict parametrize so it matches the soft test; both paths now assert that `float("Infinity")` → `inf` is rejected by the `math.isfinite` guard.

- **`_resolve_key` section in the spec doc** — keeping as-is. The rule it cites (`CLAUDE.md`: *"implementation choices … belong in pvl-core's own implementor docs or code comments, not in a spec doc"*) governs `docs/specs/`, the wire-format interop specs. This design doc lives under `docs/superpowers/specs/`, which is pvl-core's own implementor design-doc area (alongside the apps-helpers and logging-conformance design docs) — i.e. exactly the "own implementor docs" the rule points implementation detail *toward*, not the `docs/specs/` it keeps it out of. So the `_resolve_key` design rationale is correctly placed.

The other notes in the review were explicitly marked non-issues (`_reject` `%s` format is safe; `_check_bounds` return-type imprecision is mypy-clean; float-text rejection coverage adequate) — no change needed.

pvliesdonk:
Dispositions for the four findings (pushed in 327ca8e + 1bb20da):

- **Finding 1 — PEP 515 underscore acceptance.** Decided to **accept** (keep `int()`/`float()` semantics) rather than reject: rejecting only underscores while still accepting `1e3`/`+5`/non-ASCII digits would be incoherent, and readable large numbers (`10_000_000`) help downstream byte/size limits. The accept-set is now an explicit, regression-protected contract — documented in both docstrings + the design doc, and pinned by `test_accepts_pep515_underscores` (int + float) and `test_accepts_scientific_notation`.
- **Finding 2 — missing soft above-maximum int test.** Added `TestEnvInt.test_above_maximum_soft_warns_and_defaults`.
- **Finding 3 — breaking-change marker.** Maintainer decision: **keep `feat:` (minor bump)**. The old bare `ValueError` / silent-accept-out-of-range was never a documented contract, and #159 framed this strict parsing as desirable hardening rather than a breaking redesign; the coordinated server family has no `PORT=0` consumer. The behavior change is documented in the PR body, the `from_env` docstring (`Raises:`), and the design doc.
- **Finding 4 — `# TODO(#181)` comment in `_debug.py`.** **Declined.** `_debug.py` isn't in this PR's diff (adding a comment would be scope creep onto an unmodified file), and a `#181`-style forward-reference comment is the kind of rot this repo's contributor guidance avoids — the convergence is tracked in issue #181, not in a code comment.

Separately, while addressing these I also documented + pinned the trusted-default contract (soft mode returns the developer's `default` as-is; `minimum`/`maximum` validate only the operator's env value) via `test_soft_default_is_not_bounds_checked`, since several review angles noted it was correct-but-implicit.

PR #211, opened 2026-06-29
feat: add domain_env_suffixes() for decomposed-config drift checks

### LINKED ISSUE #210: Add domain_env_suffixes(): env-surface scanner for decomposed (sub-config) domain configs
## Summary

`server_config_env_suffixes()` lets a consumer (e.g. the template's config-wizard drift gate) measure the env surface of `ServerConfig`. There is no equivalent for a **domain** config that delegates its env reads into composed sub-config dataclasses, so a drift gate computing the domain surface by AST-scanning a single `from_env` undercounts.

## Observed `[verified]`

The template gate `tests/test_config_wizard_drift.py` derives the domain surface from `inspect.getsource(ProjectConfig.from_env)`, matching only `env`/`env_int`/`env_float` calls in that one function's source. A downstream that splits config into sub-config *sections* — each a frozen dataclass with its own `from_env`, composed into `ProjectConfig` via `field(default_factory=...)` — has its real reads in the sub-config sources, invisible to that scan. The coverage half of the gate then reports an empty domain surface and passes vacuously.

This is a real downstream shape: `markdown-vault-mcp` decomposed its config into a `config_sections/` package (its issue #579) and the v2.7.0 gate cannot see it.

## Need

A public counterpart to `server_config_env_suffixes()` that, given a domain config class, returns the literal env suffixes its `from_env` reads **including those delegated to composed sub-config sections**, excluding the `ServerConfig` field (already covered by `server_config_env_suffixes`). A flat config (no sub-config fields) must behave exactly as today.

## Scope boundary

Adds the scanner only; does not change `ServerConfig` or the existing suffix function. Whether/how the template gate adopts it is a separate template change.

pvliesdonk:
Closes #210.

## What

Adds a public `domain_env_suffixes(config_cls) -> frozenset[str]` to `_config.py` — the domain-side sibling of `server_config_env_suffixes()`. It AST-scans a config's `from_env` for literal `env`/`env_int`/`env_float` reads and recurses into sub-config dataclass fields (resolving types via `typing.get_type_hints`, unwrapping one level of `typing.get_args` so `Optional[Section]` / `list[Section]` are followed), excluding the composed `ServerConfig` field. Depth-first with a visited-set, so cycles and a type shared by two fields terminate/dedupe.

## Why

`server_config_env_suffixes()` lets a config-wizard drift gate measure the `ServerConfig` env surface, but a downstream that decomposes its config into sub-config *sections* (each with its own `from_env`) had no way to be measured — a single-function AST scan of `ProjectConfig.from_env` undercounts and the gate's coverage check passes vacuously. This gives decomposed configs a first-class, gate-legible surface. (Context: `markdown-vault-mcp` split its config into a `config_sections/` package; the template's v2.7.0 drift gate can't see it.)

## Design notes

- **Module-scope contract.** Real domain configs are module-level, where `get_type_hints` resolves their annotations (including `Optional[...]`). A resolution failure (broken/forward-ref annotation, or a non-module-scope config) **propagates** with a class-named message rather than silently yielding an incomplete set — appropriate for a drift gate.
- **Input + failure-path hardening.** Non-dataclass input raises `TypeError`; unreadable `from_env` source raises `OSError`; unresolvable hints raise `NameError` — each re-raised naming the offending class. Scan-only stdlib (`ast`/`inspect`/`textwrap`) is imported lazily inside the function so every consumer of `_config.py` doesn't pay for it.
- **Known limits (documented):** only unqualified `env(prefix, "LITERAL")` reads are seen (a renamed import / attribute-form / variable / keyword suffix is invisible); only one level of `get_args` is expanded (`list[Optional[Section]]` is not scanned).

## Tests

`tests/test_config.py::TestDomainEnvSuffixes` — flat (exact surface, all three read funcs), sub-config recursion + `ServerConfig` exclusion, `Optional[Sub]` via `get_args`, `list[Sub]` traversal, module-scope cycle (visited-set), same-type dedup, field-without-`from_env`, `ServerConfig`-as-root, and the three failure paths (non-dataclass → `TypeError`, source-unavailable → `OSError`, unresolvable hint → `NameError`). The `get_args` and `ServerConfig`-exclusion guards are mutation-checked. Full suite: 503 passing; `mypy src/` clean.

### DIFF
--- a/src/fastmcp_pvl_core/__init__.py
+++ b/src/fastmcp_pvl_core/__init__.py
@@ -22,7 +22,12 @@ from ._authorization import (
     parse_claim_grants,
 )
 from ._cli import make_serve_parser, normalise_http_path
-from ._config import ServerConfig, Transport, server_config_env_suffixes
+from ._config import (
+    ServerConfig,
+    Transport,
+    domain_env_suffixes,
+    server_config_env_suffixes,
+)
 from ._debug import maybe_start_debugpy
 from ._env import (
     env,
@@ -73,6 +78,7 @@ __all__ = [
     "client_supports_apps",
     "compute_app_domain",
     "configure_logging_from_env",
+    "domain_env_suffixes",
     "env",
     "env_float",
     "env_int",
--- a/src/fastmcp_pvl_core/_config.py
+++ b/src/fastmcp_pvl_core/_config.py
@@ -228,9 +228,12 @@ def domain_env_suffixes(config_cls: type) -> frozenset[str]:
     Recursion is depth-first with a visited-set, so nested sections and
     reference cycles terminate. Only literal-string suffixes are seen — a
     suffix built from a variable or passed by keyword is invisible; keep reads
-    in the ``env(prefix, "LITERAL")`` form. Sub-config types must be resolvable
-    via :func:`typing.get_type_hints` (importable in the config module's
-    namespace), which matters under ``from __future__ import annotations``.
+    in the ``env(prefix, "LITERAL")`` form. Sub-config types are discovered via
+    two mechanisms: (1) primary — resolved annotations from
+    :func:`typing.get_type_hints` (works when the config is defined at module
+    scope); (2) fallback — a field's ``default_factory`` when the annotation
+    cannot be resolved (e.g. a config defined inside a local scope under
+    ``from __future__ import annotations``).
 
     Args:
         config_cls: The domain config dataclass (its ``from_env`` classmethod is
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Thanks — addressed in `e517cdb`:

- **[Gap] instance-vs-class guard** — fixed. The guard now requires `isinstance(config_cls, type)` in addition to `is_dataclass`, so a passed instance raises `TypeError` rather than silently scanning. Added `test_dataclass_instance_raises_typeerror`.
- **[Nit] `TypeError`→`OSError` re-raise** — fixed. The `(OSError, TypeError)` handler now re-raises via `type(exc)(...)`, preserving the original exception type with class context (a type error no longer masquerades as I/O).

Declined, with rationale:

- **[Nit] `ServerConfig` not added to `visited`** — left as-is. The early return and the field-loop `candidate is not ServerConfig` guard are intentional defense-in-depth; the repeated `_visit(ServerConfig)` calls are immediate no-ops with no measurable cost, and adding it to `visited` would make the two guards' responsibilities less clear, not more.
- **[Observation] `hints.get(f.name, f.type)` fallback** — left as-is. For a well-formed dataclass `get_type_hints` always contains every field, and the `isinstance(candidate, type)` check filters the string fallback safely, so it's harmless and not worth a coverage pragma.

PR #219, opened 2026-07-04
feat(transfer): SSRF-hardened fetch_url primitive (#214)

### LINKED ISSUE #214: feat(transfer): fetch_url — SSRF-hardened URL→bytes primitive (ADR 0001 §11 #1)
Implements issue #1 of the ADR 0001 decomposition (§11). **Spec: `docs/adr/0001-transfer-lift.md` §4, §5, §8** (PR #213). Parent study: #212.

## Deliverable
A standalone, SSRF-hardened URL→bytes primitive `fetch_url`, exported from the package `__init__`, living in a new `src/fastmcp_pvl_core/_transfer/` package (`_transfer/fetch.py`). It is usable directly, independent of the transfer feature — no token store, no routes.

Lift **as-is** from `markdown-vault-mcp`'s `_server_tools/writer.py` (the mature copy; see ADR §8):
- `_resolve_pinned_ip` + `_ip_is_blocked` — DNS-rebind-safe SSRF guard (validate every resolved address, pin the validated IP into the connection, fail closed)
- http/https scheme allowlist (fixed shape — not configurable)
- blocked-hostname set; streaming size-cap loop; Host header + TLS SNI preservation with pinned-IP dial; redirects disabled
- **userinfo + query-string redaction across all emit-paths — log lines AND exception messages** (ADR §8: #122 shipped the log path but missed the exception path)

Operator config (env, per ADR §7): `TRANSFER_FETCH_MAX_BYTES`, `TRANSFER_FETCH_TIMEOUT_S`.

## Acceptance
- `fetch_url` exported in `__init__` `__all__`.
- Tests cover: blocked ranges (private/loopback/link-local/unspecified/reserved/multicast, v4+v6), DNS-rebind pin, non-http(s) scheme rejected, redirects not followed, size cap aborts early, and **no credential leakage in either logs or exception text**.

## Scope boundary
NOT in this issue: the token store, the `/transfer` routes, the link tools, base64 decode. Those are separate ADR §11 issues.

## Housekeeping (ADR §2.5/§10.4/§11)
Creating `_transfer/` is the natural moment to confirm the residual `_file_exchange/` is gone. It is an untracked empty dir (nothing to remove from git). Verify: `git ls-files | grep _file_exchange` returns nothing, and `find src -type d -name _file_exchange` is empty (rmdir if the empty dir is present locally).

## Dependencies
Blocked on ADR #213 merging (the spec). No code dependency on other decomposition issues; may land with #2.

pvliesdonk:
Implements **#214** (first slice of the ADR 0001 decomposition, §11 issue #1). Closes #214.

## What

Introduces the `_transfer/` package with a standalone, SSRF-hardened URL→bytes primitive **`fetch_url`** (+ `FetchResult`), exported from the package `__init__`. It is usable directly — no token store, no routes — and will be consumed by the later transfer ingest plane. The SSRF algorithm is lifted from `markdown-vault-mcp`'s proven fetch per ADR §8, then hardened further during review.

Also removes the empty untracked `_file_exchange/` residue (ADR §2.5/§10.4).

## Hardening (all tested as behavior contracts)

- **Address validation + DNS-rebind pin** — resolves the host, rejects if *any* resolved address is non-public, and pins the validated IP into the connection (dials the IP literal; real hostname rides only in `Host` + TLS SNI), closing the TOCTOU window. Fails **closed** on every resolution-error class (`OSError` + IDNA `UnicodeError`).
- **Permit-list block predicate** — `is_multicast or not is_global`, which structurally covers private/loopback/link-local/unspecified/reserved/benchmarking/**CGNAT (100.64/10)** on v4, v6, and IPv4-mapped forms, plus explicit **NAT64 (`64:ff9b::/96`)** embedded-IPv4 recursion (the one embedding form `is_global` doesn't decode).
- **No redirects** — every 3xx refused up front (`is_redirect`), never followed.
- **No ambient environment** — `trust_env=False` so `HTTP(S)_PROXY` / `.netrc` can't divert past the pinned IP.
- **Streaming size cap** — aborts mid-stream before an oversize body materialises.
- **Credential hygiene** — embedded userinfo dropped from the dialled request; every URL the primitive itself emits (log line + all raised messages, incl. the `HTTPStatusError` re-raised `from None`) redacted of userinfo + query via `_redact_url`.
- **Input validation** — `max_bytes > 0`, `timeout_s > 0`.

## Design for testability

Security-critical SSRF validation (`_ip_is_blocked`, `_resolve_pinned_ip` with an injected `_resolve_addresses` seam) is a pure, directly-unit-tested plane; HTTP mechanics are driven through an injected `httpx.MockTransport`. **63 contract tests**, green on Python 3.10–3.13; `ruff`/`mypy` clean; full suite (567) green.

## Review

Design brainstormed and approved (ADR 0001, PR #213, merged). Ran the local preflight circus (5 core lenses + code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer) to convergence — 7 rounds, ending fully clean. Every finding was real and fixed (CGNAT bypass, chained-cause credential leak, `trust_env` proxy bypass, NAT64 bypass, IPv6 Host-header bracketing, redaction coverage on all emit-paths). All version-dependent `ipaddress`/`httpx` claims were verified empirically across 3.10–3.13.

Docs-note: no dependency change (httpx is already core); the pre-existing `uv.lock` self-version drift (4.2.0→4.3.0) is unrelated and kept out of this PR.

### DIFF
--- a/src/fastmcp_pvl_core/__init__.py
+++ b/src/fastmcp_pvl_core/__init__.py
@@ -53,12 +53,14 @@ from ._server_info import (
     register_server_info_tool,
 )
 from ._subject import get_claims, get_subject
+from ._transfer import FetchResult, fetch_url
 
 __version__ = "4.3.0"  # PSR overrides at build time
 
 __all__ = [
     "AuthMode",
     "ConfigurationError",
+    "FetchResult",
     "IconSpec",
     "SecretMaskFilter",
     "ServerConfig",
@@ -82,6 +84,7 @@ __all__ = [
     "env",
     "env_float",
     "env_int",
+    "fetch_url",
     "get_claims",
     "get_subject",
     "load_acl",
--- a/src/fastmcp_pvl_core/_transfer/__init__.py
+++ b/src/fastmcp_pvl_core/_transfer/__init__.py
@@ -1,0 +1,15 @@
+"""In-process transfer/ingest mechanics shared across the ``*-mcp`` family.
+
+Implements ADR 0001 (``docs/adr/0001-transfer-lift.md``): SSRF-hardened URL
+fetch, size-capped base64 decode, a one-time capability-link token store, and
+the ``/transfer`` route framework. This first module ships the standalone
+``fetch_url`` primitive (§11 issue #1); the rest land in sibling issues.
+
+Intra-package imports stay relative so a fold-in is a directory rename.
+"""
+
+from __future__ import annotations
+
+from .fetch import FetchResult, fetch_url
+
+__all__ = ["FetchResult", "fetch_url"]
--- a/src/fastmcp_pvl_core/_transfer/fetch.py
+++ b/src/fastmcp_pvl_core/_transfer/fetch.py
@@ -15,8 +15,8 @@ Hardening (lifted from ``markdown-vault-mcp``'s proven fetch, ADR §8):
   in the ``Host`` header and TLS SNI). This closes the TOCTOU window between
   validation and connect. Fails **closed** on any resolution error.
 - **No redirects** — a 3xx response is refused (never followed to a possibly
-  internal target). The explicit ``is_redirect`` check refuses every 3xx up
-  front with a purpose-specific error, rather than letting a redirect surface
+  internal target). An explicit ``300 <= status < 400`` check refuses every 3xx
+  up front with a purpose-specific error, rather than letting a redirect surface
   later as httpx's generic status error.
 - **No ambient environment** — the client runs with ``trust_env=False`` so an
   ``HTTP(S)_PROXY`` env var cannot divert the request past the pinned IP and
@@ -254,7 +254,14 @@ async def fetch_url(
     hostname = parsed.hostname
     if not hostname:
         raise ValueError("The fetch URL has no host.")
-    port = parsed.port or _SCHEME_DEFAULT_PORTS[parsed.scheme]
+    # `is not None` (not truthiness): an explicit port 0 must be kept, not
+    # silently coerced to the scheme default.
+    explicit_port = parsed.port
+    port = (
+        explicit_port
+        if explicit_port is not None
+        else _SCHEME_DEFAULT_PORTS[parsed.scheme]
+    )
 
     # Resolve + validate now and pin the resulting IP into the connection: the
     # address validated here is the one actually dialled. Fails closed.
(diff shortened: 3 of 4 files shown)


pvliesdonk:
Addressed both findings from the review pass (commit `ed4b048`):

**2. Explicit port `0` coerced to default — fixed (real bug).** Switched the port logic from truthiness to `is not None` (three sites), so `http://host:0/` is now dialled on port 0 rather than silently coerced to the scheme default. Regression test `test_fetch_url_keeps_explicit_port_zero` asserts port 0 is kept in both the dial and the `Host` header (mutation-verified: reverting to truthiness fails it).

**1. Redirect refusal — the reported gap does not exist, but I made the code self-evident anyway.** The claim that `is_redirect` is `True` only for a 3xx *with* a `Location` header is not accurate for the supported httpx range. On httpx 0.28.1 (and the `>=0.27` floor), `Response.is_redirect` is `codes.is_redirect(status_code)`, whose body is literally `return 300 <= value <= 399` — purely status-based, **not** `Location`-gated. Verified empirically: `httpx.Response(304).is_redirect` and `httpx.Response(302)` *without* a `Location` header both return `True`. (The `Location`-gated predicate is the separate `has_redirect_location`.) So the prior `if response.is_redirect:` already refused every 3xx, including a `Location`-less 300/304 — no 3xx was returned as a successful `FetchResult`.

That said, `is_redirect`'s semantics are clearly easy to misread, so I replaced it with the behaviourally identical but self-evident `if 300 <= response.status_code < 400:` and added `test_fetch_url_refuses_3xx_without_location` (parametrized over `300/302/304/307/308` with **no** `Location` header). Mutation-verified: reverting the check to the `Location`-gated `has_redirect_location` makes all five cases fail — locking that the refusal is not `Location`-dependent.

SSRF-critical paths (`_ip_is_blocked`, `_resolve_pinned_ip`, `trust_env=False`, IP pinning, redaction) are unchanged. 69 tests, green on 3.10–3.13.

pvliesdonk:
Re: the round-2 note that `httpx`'s `is_redirect` is a 5-element set `{301,302,303,307,308}` excluding 300/304 — that is not correct for the supported `httpx` (`>=0.27`, resolved 0.28.1). Verified from the installed source:

```python
>>> import inspect, httpx; httpx.__version__
'0.28.1'
>>> print(inspect.getsource(httpx.codes.is_redirect))
    @classmethod
    def is_redirect(cls, value: int) -> bool:
        """Returns `True` for 3xx status codes, `False` otherwise."""
        return 300 <= value <= 399
>>> [httpx.Response(c).is_redirect for c in (300, 304)]
[True, True]
```

`Response.is_redirect` delegates to `codes.is_redirect`, which is a `300 <= value <= 399` **range**, not set membership — so `300` and `304` return `True`. (The 5-element, `Location`-gated predicate is the *separate* `has_redirect_location`.) The prior `if response.is_redirect:` therefore already refused every 3xx including 300/304; there was no gap, and the `test_fetch_url_refuses_3xx_without_location` parametrization (300/302/304/307/308, no `Location`) passes against the range check and would also pass against `is_redirect`.

So the "equivalent" comment is accurate, and I'm keeping the explicit `300 <= status < 400` check (it's clearer and version-independent). No code change needed — flagging the underlying `httpx`-semantics claim as incorrect rather than adjusting the code around it.

All checks are green (tests 3.10–3.13, CodeQL, claude-review), no unresolved threads.

Issue #217, opened 2026-07-04
feat(transfer): TransferSink/TransferValidator + make_transfer_handler route (ADR 0001 §11 #4)


pvliesdonk:
Implements issue #4 of the ADR 0001 decomposition (§11). **Spec: `docs/adr/0001-transfer-lift.md` §3, §4, §5** (PR #213). Parent study: #212.

## Deliverable
The domain seam plus the HTTP handler:
- `_transfer/sink.py` — the `TransferSink` Protocol (`read(handle) -> (body, media_type, filename)`, `write(handle, body) -> payload`) and the `TransferValidator` type (caller ref + kind → validated opaque handle). These are the **only** things a downstream implements.
- `_transfer/routes.py` — `make_transfer_handler(store, sink)`: GET = download (`sink.read`), POST/PUT = upload (`sink.write`), receive-side streaming size-cap, RFC 6266 Content-Disposition, burn-on-success / release-on-failure.

The sink is **byte-oriented and bounded by the size caps** (ADR §3); constant-memory streaming is deliberately deferred and out of scope here.

## Acceptance
- Tests: download serves bytes then burns the token; upload commits via `sink.write` then burns; a handler failure calls `store.release` so the link survives; over-cap upload → 413; unknown/expired/consumed token → 404; a validator rejection surfaces correctly.

## Scope boundary
NOT in this issue: `register_transfer_routes`, the two link tools, `TransferConfig`, the `__init__` exports (those are #218).

## Dependencies
Depends on **#216** (`TransferStore`). Blocked on ADR #213 merging.

— 🤖 _Automated post by Claude Code (agent) via the account owner's GitHub token; agent analysis/proposal, not a personal directive from the account owner._

PR #221, opened 2026-07-05
feat(transfer): KV-backed TransferStore — one-time token store (#216)

### LINKED ISSUE #216: feat(transfer): KV-backed TransferStore — one-time token store (ADR 0001 §11 #3)
Implements issue #3 of the ADR 0001 decomposition (§11). **Spec: `docs/adr/0001-transfer-lift.md` §6** (the store's full spec) — also §2.3. PR #213. Parent study: #212.

## Deliverable
`TransferStore` in `src/fastmcp_pvl_core/_transfer/store.py`, backed by `build_kv_store(config, namespace=\"transfer\")` (whose docstring already names this consumer). Store only — no routes, no tools.

Per ADR §6:
- Token carries an **opaque `sink_handle`** the store never interprets, plus kind/expiry/caps.
- `available → in_flight → consumed` machine: `claim(token, kind)` marks in_flight with a short lease; `release(token)` reverts in_flight→available on transient failure (**the one-time link survives**); `complete(token)` burns.
- Expiry via the KV entry TTL (replaces a sweep loop); an abandoned reservation auto-frees once its lease/TTL lapses.
- Correctness via an **in-process `asyncio.Lock`** around the read-modify-write (single-process deployment → exact one-time semantics; §6.3).
- **release-on-failure is required; delete-on-claim is explicitly rejected** (ADR §6.1).

## Acceptance
- Tests: claim→complete burns; claim→release keeps the link claimable again; lease reclaim of a crashed (never-released) in_flight token; TTL expiry removes the token; wrong-kind claim rejected; a mutation test confirming delete-on-claim behaviour is NOT present (release-on-failure holds).

## Scope boundary
NOT in this issue: `/transfer` routes, link tools, the sink/validator protocols. Cross-replica CAS single-use is explicitly out of scope (ADR §6.3: not planned; TTL is the security bound).

## Dependencies
Blocked on ADR #213 merging. Independent of #214/#215; **#217** (sink + handler) and **#218** (register) depend on this.

pvliesdonk:
Implements **ADR 0001 §6 / §11 issue #3** (`docs/adr/0001-transfer-lift.md`): the one-time capability-link token store that the sink/route layer (#217/#218) will build on.

Closes #216.

## What lands

`TransferStore` in `src/fastmcp_pvl_core/_transfer/store.py`, backed by `build_kv_store(config, namespace="transfer")`, running the `available → in_flight → consumed` state machine:

- **`mint`/`claim`/`release`/`complete`** serialised by one in-process `asyncio.Lock` — exact single-use under the single-process deployment (ADR §6.3). `mint` is lock-free (a fresh 32-byte secret cannot collide).
- **Release-on-failure preserved (ADR §6.1)** — a failed reservation reverts `in_flight → available` so the one-time link survives a transient serve failure; delete-on-claim is rejected. A late `release` after a successful `complete` never resurrects the burned token.
- **Lease-based reclaim** of a crashed handler's `in_flight` token, checked read-side in `claim` (no sweep loop).
- **TTL is the security bound** — the KV entry TTL governs lifetime; every mutating op re-puts with the *remaining* TTL so the link never becomes immortal. (Empirically, drift is net-negative: ~−0.06s over 500 claim/release cycles — links expire slightly sooner, never later.)
- **Opaque `sink_handle`** (typed `object`, so consumers must narrow) + `caps` the store never interprets; only `kind` is matched on claim. `from_config` pins `namespace="transfer"` as a pvl-core shape decision.
- The persisted record is a `TypedDict` with a `Literal` status, so mypy checks every status comparison/assignment (a mistyped state fails the type check) while the value stays a plain JSON-serialisable dict.

Store only — no routes/tools. Kept internal (importable via `._transfer.store`) until the sink/route layer consumes it; not added to any public `__all__`.

## Tests

26 contract tests cover the state machine, release-on-failure, late-release-not-resurrect, lease reclaim vs live lease, TTL expiry, TTL preservation on claim/release/complete, wrong-kind, validation, the pinned `transfer` namespace, and a **file-backed round-trip** of the opaque fields. The concurrency test uses a `_YieldingKV` wrapper that forces a real read→write interleave (a plain in-memory backend never suspends) — mutation-verified to fail with a double-claim when the lock is removed.

## Deliberately deferred

For this single-shape, internal, not-yet-consumed store: a discriminated-union / schema-version guard on the record, and a mint-time JSON-serialisability probe (the backend enforces serialisability at persist time, inside `mint`). Revisit when a second record shape lands under a live consumer.

## Follow-up (minor)

`TokenNotClaimedError`'s message says "never claimed" but also fires on the claim→release→complete misuse path (the token *was* claimed then released). Reword to "not currently claimed (never claimed, or already released)" — cosmetic, on a misuse path.

### DIFF
--- a/src/fastmcp_pvl_core/_transfer/__init__.py
+++ b/src/fastmcp_pvl_core/_transfer/__init__.py
@@ -2,9 +2,10 @@
 
 Implements ADR 0001 (``docs/adr/0001-transfer-lift.md``): SSRF-hardened URL
 fetch, size-capped base64 decode, a one-time capability-link token store, and
-the ``/transfer`` route framework. Landed so far: ``fetch_url`` (§11 issue #1)
-and ``decode_base64_capped`` (§11 issue #2); the token store and route
-framework land in sibling issues.
+the ``/transfer`` route framework. Landed so far: ``fetch_url`` (§11 issue #1),
+``decode_base64_capped`` (§11 issue #2), and ``TransferStore`` in
+``store.py`` (§11 issue #3, kept internal until the sink/route layer consumes
+it); the sink protocols and route framework land in sibling issues.
 
 Intra-package imports stay relative so a fold-in is a directory rename.
 """
--- a/src/fastmcp_pvl_core/_transfer/store.py
+++ b/src/fastmcp_pvl_core/_transfer/store.py
@@ -53,6 +53,9 @@ _TOKEN_COLLECTION = "tokens"
 _TOKEN_BYTES = 32
 """Entropy of a minted token — an unguessable 32-byte URL-safe secret."""
 
+_FENCE_BYTES = 8
+"""Entropy of a per-claim fence — a fresh reservation id each :meth:`claim`."""
+
 _NAMESPACE = "transfer"
 """The KV namespace pvl-core pins for the transfer subsystem (a shape decision:
 downstream does not choose the keyspace)."""
@@ -84,6 +87,11 @@ class _TokenRecord(TypedDict):
     sink_handle: object
     caps: Mapping[str, Any]
     lease_expires_at: float | None
+    # A fresh id stamped on each claim (the fencing token). ``release`` and
+    # ``complete`` only act when the caller's fence matches this — so a stale
+    # holder whose lease was reclaimed cannot mutate the new holder's
+    # reservation. ``None`` while ``available``/``consumed``.
+    fence: str | None
 
 
 class TransferTokenError(Exception):
@@ -107,7 +115,11 @@ class TokenInFlightError(TransferTokenError):
 
 
 class TokenNotClaimedError(TransferTokenError):
-    """:meth:`TransferStore.complete` was called on a token that was never claimed."""
+    """The token is not currently claimed.
+
+    Raised by :meth:`TransferStore.complete` when the token is ``available`` —
+    never claimed, or already released.
+    """
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Fixed the lease-fencing gap in `3b02628` (as decided with the maintainer):

- `claim()` now stamps a fresh per-reservation `fence` (`secrets.token_urlsafe(8)`) into the record and returns it on `TransferToken`.
- `release(token, fence)` and `complete(token, fence)` no-op when the record's current fence doesn't match — so a slow-but-alive handler whose lease was reclaimed by a second `claim()` can no longer revert or burn the new holder's live reservation. Single-process one-time-use is now exact including the slow-handler case, not just the crashed-handler case.
- Random fence (not a monotonic counter) is deliberate: a counter would reset to 0 on restart while records survive in KV, letting a post-restart `claim` regenerate a value a pre-restart stale holder still carries — reintroducing the collision. A 64-bit random fence has no restart-reset failure mode.

New regression test `test_superseded_holder_cannot_mutate_reclaimed_reservation` was mutation-verified to fail if **either** the `release` or the `complete` fence guard is removed (the first version of the test missed the `complete` guard because both A- and B-burns end in `consumed`; it was strengthened to assert the token stays B's *live in-flight* claim after A's stale calls).

Also folded in your earlier note: `TokenNotClaimedError`'s message was reworded from "never claimed" to "not currently claimed (available — never claimed, or already released)" since it also fires on the claim→release→complete path.

Two minor, deliberately-deferred hardening items (both surfaced sub-threshold locally; neither bites while the store is unconsumed, so they'll land naturally with the sink/route layer in #217, where the real call sites exist):
- `release`/`complete` take two same-typed positional `str` args (`token`, `fence`) — a swapped call would silently no-op. Worth making keyword-only or accepting the `TransferToken` handle once #217 defines the call convention.
- `release()`-on-expired-token no-op is the one untested branch (its `complete()` sibling is covered) — a two-line mirror test would take coverage to 100%.

Issue #223, opened 2026-07-05
feat(transfer): grace-settle completion (TTL reduction) instead of hard burn; keep burn as explicit option


pvliesdonk:
## Observed problem

The `/transfer/{token}` download handler (#217) must call `store.complete` (which currently **burns** the token → `consumed`) *before* returning the Starlette `Response`, because the body is transmitted to the client only after the handler returns. So a client stall/disconnect during download transmission spends the one-time link with nothing delivered — the client cannot retry. This is the exact hazard `markdown-vault-mcp`'s implementation hit in practice, and it is what ADR §6.1 invokes ("a download stalls … forcing a re-mint") to justify release-on-failure — yet the buffered-Response model reintroduces it on the transmission phase.

`[verified: traced in routes.py _download — complete() precedes the Response return]`

## Decided resolution (maintainer-directed)

Lean on the TTL — which ADR §6.3 already names as **the** security-relevant bound, with strict single-use as *hygiene* — rather than a hard burn:

- **`complete(token, fence)` becomes grace-settle**: `status → available`, `fence → None`, `TTL → min(remaining, GRACE_S)`. The `min` (never extend) pins the absolute expiry at *first-complete + grace*, so retries within the grace window re-serve without sliding the bound. After grace, ordinary KV-TTL expiry removes the token — no sweep.
- **`release(token, fence)` stays distinct**: `→ available` with the full remaining TTL (long retry window). complete and release both revert to available; they differ only in remaining life.
- **A stalled/redisconnected download simply retries within grace** — the link is no longer stranded.
- **Keep `consumed` as an explicit optional hard burn**: retain the `consumed` state + `TokenAlreadyConsumedError`, exposed as an explicit `burn(token, fence)` for a caller that genuinely wants strict one-shot (e.g. a sensitive one-time secret where even a grace-window replay is unacceptable).

New operator env var (wired in #218): `TRANSFER_GRACE_TTL_S`.

## Deliverables

- Amend `docs/adr/0001-transfer-lift.md` §6 to the grace-settle state machine (default success = grace-settle; explicit `burn` = strict one-shot), noting it supersedes the original hard-burn after the burn-before-transmit / MVM stranding realization.
- `store.py`: add a `grace_seconds` constructor arg; `complete` → grace-settle; add `burn`; keep `consumed`/`TokenAlreadyConsumedError`.
[...]

Issue #218, opened 2026-07-04
feat(transfer): register_transfer_routes + link tools + TransferConfig + exports (ADR 0001 §11 #5)


pvliesdonk:
Implements issue #5 of the ADR 0001 decomposition (§11) — the piece that ties the framework together. **Spec: `docs/adr/0001-transfer-lift.md` §3, §5, §7** (PR #213). Parent study: #212.

## Deliverable
- `_transfer/register.py` — `register_transfer_routes(mcp, config, *, sink, validate)`: builds one shared `TransferStore`, wires the `/transfer/{token}` route (via the #217 handler) and the two link tools `create_download_link` / `create_upload_link`, applies the TTL clamp and the `base_url`-required guard. Owns route/tool **shape**; `sink` + `validate` are the only hook kwargs (no override kwargs — ADR §7/§10.2).
- `_transfer/config.py` — `TransferConfig` with `from_env(prefix)` (`TRANSFER_TTL_DEFAULT_S`, `TRANSFER_TTL_MAX_S`, `TRANSFER_MAX_UPLOAD_BYTES`; fetch knobs land with #214), registered in the `domain_env_suffixes()` drift gate.
- `__init__` exports for `register_transfer_routes` (and `TransferConfig`, `TransferSink`, `TransferValidator` as the public surface).

## Acceptance
- End-to-end tests: mint a download link → redeem once → 404 on second redeem; mint an upload link → commit → 404 on second; TTL clamp (unset → default, over-max → clamped); `base_url` unset → mint raises; `TransferConfig.from_env` parsing + presence in the `domain_env_suffixes()` drift gate.

## Scope boundary
NOT in this issue: downstream sink implementations / migrations (`markdown-vault-mcp`, `image-generation-mcp`) — those are separate trackers in their own repos (ADR §9).

## Dependencies
Depends on **#216** (`TransferStore`) and **#217** (sink/validator + handler). Blocked on ADR #213 merging.

— 🤖 _Automated post by Claude Code (agent) via the account owner's GitHub token; agent analysis/proposal, not a personal directive from the account owner._

PR #225, opened 2026-07-05
feat(transfer): TransferSink seam + make_transfer_handler route (#217)

### LINKED ISSUE #217: feat(transfer): TransferSink/TransferValidator + make_transfer_handler route (ADR 0001 §11 #4)

pvliesdonk:
Implements ADR 0001 §3/§4/§5/§8 / §11 issue #4 on top of the grace-settle store (#223/#224): the domain seam and the `/transfer/{token}` HTTP handler. Closes #217.

## What lands

- **`sink.py`** — the `TransferSink` Protocol (`read(handle) -> TransferReadResult`, `write(handle, body) -> payload`) + the `TransferValidator` type. `TransferReadResult` is a `NamedTuple` (not a bare 3-tuple) so `media_type`/`filename` — both `str` — are named rather than swap-prone positionals. These are the **only** things a downstream implements.
- **`routes.py`** — `make_transfer_handler(store, sink, *, max_upload_bytes)`, a Starlette handler. GET downloads via `sink.read` + an RFC 6266 `Content-Disposition`; POST/PUT upload via a receive-side chunked size-cap read + `sink.write`. On success `store.complete` (**grace-settle** — a served-but-stalled transfer can retry within the grace window instead of being stranded); on any failure incl. cancellation `store.release`. Claim errors map to HTTP: unknown/expired/consumed/wrong-kind → 404, in-flight → 409, over-cap → 413.

## Hardening

- **`Connection: close`** on every response that abandons an unread request body (413 over-cap, 404/409 upload claim errors, and the 405 unserved-method branch) — an undrained body on a keep-alive socket desyncs the next request. `_CLOSE_CONN` is a read-only `MappingProxyType`.
- **`_release_quietly`** — releases on the failure path, logging+swallowing an *ordinary* release error (a `BaseException`/cancellation still propagates) so the original transfer error is never masked. It logs only the release-error **class**, never a traceback/message — some KV backends embed the token-derived key in error text, and the token is a secret.
- A module logger (the Starlette route bypasses the MCP logging middleware): logs claim-rejection class, over-cap, and release failure — never the token.
- **Content-Disposition** strips CR/LF + quoted-string breakers (no header injection); RFC 5987 `filename*` for non-ASCII.

`TransferValidator` is defined here but invoked at link-creation by the link tools (#218), not by the handler. Kept internal until #218 registers the route and wires exports.

## Tests

19 handler contract tests over a real Starlette route via httpx ASGI — grace-settle reclaim-within-grace + expiry-after-grace, upload commit + grace retry, sink read/write failure release+retry, a **multi-chunk over-cap** test (pins the cross-chunk cap accumulation), `Connection: close` on every undrained-body branch, a release-failure-does-not-mask-the-original test, in-flight 409, 405, non-ASCII RFC 5987, and CRLF header-injection defense. The size-cap, release, burn-vs-grace-settle, 405-close, and release-mask guards are all mutation-verified.

## Note

This is the rebuild of the parked #217 handler onto grace-settle: the download burn-before-transmit gap the original review surfaced is dissolved (completing no longer strands the link), and all the other first-review findings (Connection: close, multi-chunk cap test, logging, release guard, NamedTuple, drop `@runtime_checkable`) are applied here.

### DIFF
--- a/src/fastmcp_pvl_core/_transfer/__init__.py
+++ b/src/fastmcp_pvl_core/_transfer/__init__.py
@@ -1,11 +1,13 @@
 """In-process transfer/ingest mechanics shared across the ``*-mcp`` family.
 
 Implements ADR 0001 (``docs/adr/0001-transfer-lift.md``): SSRF-hardened URL
-fetch, size-capped base64 decode, a one-time capability-link token store, and
-the ``/transfer`` route framework. Landed so far: ``fetch_url`` (§11 issue #1),
-``decode_base64_capped`` (§11 issue #2), and ``TransferStore`` in
-``store.py`` (§11 issue #3, kept internal until the sink/route layer consumes
-it); the sink protocols and route framework land in sibling issues.
+fetch, size-capped base64 decode, a capability-link token store, and the
+``/transfer`` route framework. Landed so far: ``fetch_url`` (§11 issue #1),
+``decode_base64_capped`` (§11 issue #2), ``TransferStore`` in ``store.py``
+(§11 issue #3), and — in ``sink.py`` and ``routes.py`` — the ``TransferSink`` /
+``TransferValidator`` domain seam plus ``make_transfer_handler`` (§11 issue #4).
+All of these except the two exported primitives stay internal until the
+route-registration layer (§11 issue #5) wires them into the public surface.
 
 Intra-package imports stay relative so a fold-in is a directory rename.
 """
--- a/src/fastmcp_pvl_core/_transfer/routes.py
+++ b/src/fastmcp_pvl_core/_transfer/routes.py
@@ -1,0 +1,208 @@
+"""The ``/transfer/{token}`` HTTP handler (ADR 0001 §3/§8 / §11 #4).
+
+:func:`make_transfer_handler` builds the Starlette handler that drives the
+capability-link state machine over HTTP:
+
+- **GET** downloads: claim the token, ``sink.read`` the bytes, serve them with
+  an RFC 6266 ``Content-Disposition``, then ``store.complete`` (grace-settle).
+- **POST/PUT** uploads: claim the token, read the request body under a
+  size cap (aborting early past it), ``sink.write`` it, return the sink's
+  payload as JSON, then ``store.complete`` (grace-settle).
+
+The link is **settled on success** (``store.complete`` — grace-settle, so a
+served-but-stalled transfer can retry within the grace window rather than be
+stranded) and **released on any failure** (``store.release`` — a transient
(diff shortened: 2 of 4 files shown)


pvliesdonk:
Good catch on the 405 `Connection: close` reachability — confirmed empirically: with a production registration of `methods=["GET","POST","PUT"]`, a `DELETE`-with-body gets Starlette's own Router-level 405 (with `Allow`, **no** `Connection: close`) *before* the handler runs, so the handler's 405+close branch never fires for it. `HEAD` is auto-added for `GET` and does reach the branch, but is bodyless.

Fixed in `9dec612`:
- Added an `Allow: GET, POST, PUT` header to the handler's 405 (RFC 7231 §6.5.5; parity with Starlette's built-in 405).
- Documented in `make_transfer_handler`'s `Note`: for the close to cover **every** unsupported method, the route-registration layer (#218) must register the route so all methods reach this handler (omit `methods=` or pass a superset); otherwise Starlette's own 405 answers unregistered methods without closing the connection. Recorded as an explicit registration requirement for #218.
- The test comment now states honestly that it exercises the handler's own 405 branch, and that ensuring every method reaches it in production is #218's job; it also asserts the `Allow` header.

On the minor `Allow`-header nit — done in the same commit.

CI is green (tests 3.10–3.13, CodeQL, and the prior claude-review run all passed).

